PR #199, opened 2019-02-27
Androapp 1661

ferdyrod:
This pull requests enables support por hideOption and hideOptionGroup rule actions, when the rule Engine add is support to those same actions. Which mean that until the rule engine is not updated, this will continue to be unsupported.


ferdyrod:
@Balcan branch updated for merging.

PR #800, opened 2019-09-19
Bug SpinnerViewModel into dataEntryDillCallBack

MarinaCape:


Balcan:
Using autovalue now fixed the problem. Closing the request.

PR #824, opened 2019-09-25
Prem Patch

khoran906:
Applied patch stitched from the last several commits by Prem on iupui-soic fork of dhis2-android-trackercapture.


Balcan:
This is a custom modification of the android app so it won't be included in the official repository.

PR #896, opened 2019-10-14
refactor: optionSetDialog with presenter

mmmateos:
BREAKING CHANGES: OptionSetDialog with presenter used in events and datasets. Migrated to Kotlin


app/src/main/java/org/dhis2/utils/optionset/OptionSetContracts.kt
```diff
@@ -0,0 +1,28 @@
...
+        fun setLiveData(data: LiveData<PagedList<Option>>?)
+
+        fun showDialog(): Boolean
+    }
+
+    interface Presenter : AbstractActivityContracts.Presenter {
+
+        fun init(view: View, optionSet : SpinnerViewModel, textSearch : EditText)
+
+        fun init(view: View, optionSetTable: org.dhis2.data.forms.dataentry.tablefields.spinner.SpinnerViewModel, textSearch: EditText)
```
ferdyrod: In kotlin we can use importing aliases. So basically we can import SpinnerViewModel from both places as follows.

    import org.dhis2.data.forms.dataentry.fields.spinner.SpinnerViewModel
    import org.dhis2.data.forms.dataentry.tablefields.spinner.SpinnerViewModel **as TableSpinnerViewModel**

And now we can update the table init as follows
    `fun init(view: View, optionSetTable: **TableSpinnerViewModel**, textSearch: EditText)`

app/src/main/java/org/dhis2/data/user/UserComponent.java
```diff
@@ -175,4 +177,7 @@
 
     @NonNull
     ScheduledEventComponent plus(ScheduledEventModule scheduledEventModule);
+
+    @NonNull
+    OptionSetComponent plus(OptionSetModule optionSetModule);
```
Balcan: I would have added the OptionSetComponent as a subcomponent of the ServerComponent or the AppComponent as it just need D2 and the ScheduleProvider.

app/src/main/java/org/dhis2/utils/optionset/OptionSetContracts.kt
```diff
@@ -0,0 +1,29 @@
...
+class OptionSetContracts {
+
+    interface View {
+
+        fun setLiveData(data: LiveData<PagedList<Option>>?)
+
+        fun showDialog(): Boolean
+    }
+
+    interface Presenter : AbstractActivityContracts.Presenter {
```
JaimeToca: I think you don't need this interface Presenter :dancer:

PR #911, opened 2019-10-21
feat(preferences): Copy remaining preferences into the secured shared preference when app starts.

Balcan:
- Had to refactor some code in splash and login screen. Removed unnecessary methods from base activity.
- MainActivity switch to kotlin


app/src/main/java/org/dhis2/usescases/login/LoginActivity.kt
```diff
@@ -66,7 +64,7 @@ class LoginActivity : ActivityGlobalAbstract(), LoginContracts.View {
     private var qrUrl: String? = null
 
     private var testingCredentials: List<TestingCredential> = ArrayList()
-    var userManager : org.dhis2.data.server.UserManager? = null
+    var userManager: org.dhis2.data.server.UserManager? = null
```
ferdyrod: There is a reason this declaration of the userManager variable has to be explicit?

app/src/main/java/org/dhis2/usescases/splash/SplashActivity.kt
```diff
@@ -107,4 +109,13 @@ class SplashActivity : ActivityGlobalAbstract(), SplashContracts.View {
         }
     }
 
+    override fun goToNextScreen(isUserLogged: Boolean, sessionLocked: Boolean, initialSyncDone: Boolean) {
+        if (isUserLogged && initialSyncDone && !sessionLocked) {
+            startActivity(MainActivity::class.java, null, true, true, null)
+        } else if (isUserLogged && !initialSyncDone) {
+            startActivity(MainActivity::class.java, null, true, true, null)
```
ferdyrod: Since initialSyncDone can be either true or false and execute the same code, maybe we can omit one if clause and just check for isUserLogged and !sessionLocked?

app/src/main/java/org/dhis2/usescases/splash/SplashPresenter.kt
```diff
@@ -1,63 +1,55 @@
...
+                Constants.PREFS_URLS, Constants.PREFS_USERS ->{}
+                else -> preferenceProvider.setValue(it.key, it.value)
+            }
+        }
         isUserLoggedIn()
     }
 
-    override fun isUserLoggedIn() {
+    private fun isUserLoggedIn() {
         if (userManager == null)
```
ferdyrod: This is good. Just for information we can do something as 
userManager?.let {

    // userManager is not null here

} ?: //Execute when userManager is null here!

app/src/main/java/org/dhis2/usescases/login/LoginPresenter.kt
```diff
@@ -2,30 +2,33 @@ package org.dhis2.usescases.login
...
+import org.dhis2.utils.analytics.CLICK
+import org.dhis2.utils.analytics.LOGIN
+import org.dhis2.utils.analytics.SERVER_QR_SCANNER
 import org.hisp.dhis.android.core.maintenance.D2Error
 import org.hisp.dhis.android.core.maintenance.D2ErrorCode
 import org.hisp.dhis.android.core.systeminfo.SystemInfo
 import retrofit2.Response
 import timber.log.Timber
 
 class LoginPresenter(private val view: LoginContracts.View,
```
ferdyrod: We need to also update the LoginPresenter Tests with the new name of the SharedPreferences' Constants names.

app/src/main/java/org/dhis2/usescases/main/MainPresenter.kt
```diff
@@ -0,0 +1,126 @@
...
+
+class MainPresenter(
+        private val view: MainView,
+        private val d2: D2,
+        private val schedulerProvider: SchedulerProvider) {
+
+    private var compositeDisposable: CompositeDisposable? = null
+
+    fun init() {
+        this.compositeDisposable = CompositeDisposable()
```
ferdyrod: Should we initialize the disposable outside of the init method?

app/src/main/java/org/dhis2/usescases/splash/SplashActivity.kt
Balcan: second branch should point to SyncActivity

app/src/main/java/org/dhis2/usescases/main/MainPresenter.kt
Balcan: Ye

app/src/main/java/org/dhis2/usescases/login/LoginActivity.kt
```diff
@@ -305,8 +304,8 @@ class LoginActivity : ActivityGlobalAbstract(), LoginContracts.View {
         binding.serverUrlEdit.dropDownWidth = resources.displayMetrics.widthPixels
         binding.userNameEdit.dropDownWidth = resources.displayMetrics.widthPixels
 
-        urls = getListFromPreference(Constants.PREFS_URLS)
-        users = getListFromPreference(Constants.PREFS_USERS)
+        urls = presenter.preferenceProvider.getSet(Constants.PREFS_URLS, emptySet())!!.toMutableList()
```
JaimeToca: A few things :)
Views should not know about datastores, therefore we should not expose preferencesProvider as public.
The way this should go is :
view calls presenter, for ej: presenter.renderURL(), presenter fetch from preferences provider and then send the results to the view

JaimeToca: Can this !! throw a null pointer exception? careful with this hehe.

app/src/main/java/org/dhis2/usescases/main/MainPresenter.kt
```diff
@@ -0,0 +1,126 @@
...
+import org.dhis2.usescases.login.LoginActivity
+import org.dhis2.utils.Constants
+import org.dhis2.utils.filters.FilterManager
+import org.hisp.dhis.android.core.D2
+import org.hisp.dhis.android.core.user.User
+import timber.log.Timber
+
+const val DEFAULT = "default"
+
+class MainPresenter(
```
JaimeToca: This class was totally change or it was the code style :) ?

app/src/main/java/org/dhis2/usescases/login/LoginActivity.kt
```diff
@@ -104,12 +103,12 @@ class LoginActivity : ActivityGlobalAbstract(), LoginContracts.View {
             binding.userPassEdit.setText(testingEnvironment.val2())
         })
         loginViewModel.serverUrl.observe(this, Observer<String> {
-           /* Glide.with(this).load(String.format("%s/api/staticContent/logo_front", it))
-                    .transition(withCrossFade())
-                    .into(binding.logoFront)
-            Glide.with(this).load(String.format("%s/api/staticContent/logo_banner", it))
-                    .placeholder(R.drawable.ic_dhis_white)
-                    .into(binding.logoBanner)*/
+            /* Glide.with(this).load(String.format("%s/api/staticContent/logo_front", it))
```
JaimeToca: do we need this for something or we can remove it?

app/src/main/java/org/dhis2/data/prefs/PreferenceProviderImpl.kt
```diff
@@ -56,7 +80,8 @@ class PreferenceProviderImpl(val context: Context) : PreferenceProvider {
     }
 
     override fun areCredentialsSet(): Boolean {
-        return SecurePreferences.getBooleanValue(context, Constants.SECURE_CREDENTIALS, false)
+        return SecurePreferences.getBooleanValue(context, Constants.SECURE_CREDENTIALS,
```
JaimeToca: for the future (not now), this class should only contains atomic operations not validations or any others

app/src/main/java/org/dhis2/data/prefs/Preference.kt
```diff
@@ -0,0 +1,9 @@
+package org.dhis2.data.prefs
+
+class Preference {
```
JaimeToca: :100:

app/src/main/java/org/dhis2/usescases/splash/SplashPresenter.kt
```diff
@@ -1,63 +1,55 @@
...
 
-    override fun destroy() {
+    fun destroy() {
         compositeDisposable.clear()
     }
 
-    override fun init(view: SplashContracts.View) {
-        this.view = view
+    fun init() {
+        preferenceProvider.sharedPreferences().all.forEach {
```
JaimeToca: For next time :+1:  I would move this code to preferences provider o a class with only the responsability of migration one preferences to another. we should try to have presenters only with the "view logic" and fetching from data

app/src/main/java/org/dhis2/usescases/login/LoginPresenter.kt
JaimeToca: Next time the build will fail :)

PR #915, opened 2019-10-23
fix: show table correctly for any number of headers

mmmateos:
Also improve performance displaying tables

### DIFF
--- a/app/src/main/java/org/dhis2/usescases/datasets/dataSetTable/dataSetSection/DataValuePresenter.java
+++ b/app/src/main/java/org/dhis2/usescases/datasets/dataSetTable/dataSetSection/DataValuePresenter.java
@@ -488,26 +488,25 @@ public class DataValuePresenter implements DataValueContract.Presenter {
     }
 
     @Override
-    public List<List<String>> getCatOptionCombos(List<List<Pair<CategoryOption, Category>>> listCategories, int num, List<List<String>> result, List<String> current) {
-        if (num == listCategories.size()) {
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/dhis2/usescases/datasets/dataSetTable/dataSetSection/DataValuePresenter.java
```diff
@@ -513,8 +499,10 @@ public void initializeProcessor(@NonNull DataSetSectionFragment dataSetSectionFr
         for (int i = 0; i < listCategories.get(num).size(); i++) {
             if (num == 0)
                 current = new ArrayList<>();
-            if (current.size() == num + 1)
-                current.remove(current.size() - 1);
+            for (int j = current.size(); j > 0; j--) {
+                if (current.size() == num + j)
+                    current.remove(current.size() - j);
+            }
```
JaimeToca: What is "num", "i", "j" and "current" ? Can you give them a more expressive name ? ex: headerRowPositon etc..
What this expression (current.size() == num + j) means?

mmmateos: Okey, I'll try to make it more expressive :sweat_smile:

JaimeToca: no problem :+1:

PR #918, opened 2019-10-23
test: refactors MainPresenter and adds unit tests

ferdyrod:


app/src/main/java/org/dhis2/usescases/main/MainPresenter.kt
```diff
@@ -20,60 +21,55 @@ const val DEFAULT = "default"
...
         )
 
-        compositeDisposable!!.add(
-            d2.categoryModule().categoryCombos().byIsDefault().eq(true).one().get().toObservable()
+        disposable.add(
+            d2.categoryModule().categoryCombos().byIsDefault().eq(true).one().get()
                 .subscribeOn(schedulerProvider.io())
                 .subscribe(
                     { categoryCombo ->
-                        val prefs = view.abstracContext.getSharedPreferences(
```
JaimeToca: :+1:

app/src/test/java/org/dhis2/usecases/main/MainPresenterTest.kt
```diff
@@ -0,0 +1,168 @@
...
+    private val preferences: PreferenceProvider = mock()
+    private val workManger: WorkManager = mock()
+
+    @Before
+    fun setUp() {
+        presenter = MainPresenter(view, d2, schedulers, preferences, workManger)
+    }
+
+    @Test
+    fun `Should show username on activity resumed`() {
```
JaimeToca: These tests belong to the same **flow**, function and usecase which is init() the presener. Perhaps they should be compacted as one test... Let me know what you think :+1: 

`presenterMocks()
presenter.init()
//all verifies`

app/src/test/java/org/dhis2/usecases/main/MainPresenterTest.kt
```diff
@@ -0,0 +1,168 @@
...
+    }
+
+    @Test
+    fun `Should save default categoryOptionCombo uid to preferences`() {
+        presenterMocks()
+
+        presenter.init()
+
+        verify(preferences).setValue(PREF_DEFAULT_CAT_OPTION_COMBO, "uid")
+    }
```
JaimeToca: These 4 tests belong to the same flow, function and usecase which is init() the presener. Perhaps they should be compacted as one test... Let me know what you think +1

presenterMocks() 
presenter.init() //all verifies

app/src/test/java/org/dhis2/usecases/main/MainPresenterTest.kt
```diff
@@ -0,0 +1,168 @@
...
+    @Test
+    fun `Should open drawer when menu is clicked`() {
+        presenter.onMenuClick()
+
+        verify(view).openDrawer(any())
+    }
+
+    private fun presenterMocks() {
+        // UserModule
+        whenever(d2.userModule()) doReturn mock()
```
JaimeToca: I know all this mocking is huge. We'll try to return mocks from sdk dagger modules... I'll keep you updated ;)

JaimeToca:
Very nice PR @ferdyrod ! I posted a few comments. Let me know what you think ;)

cheers

app/src/test/java/org/dhis2/usecases/main/MainPresenterTest.kt
ferdyrod: I completely agree with this.

PR #923, opened 2019-10-28
refactor: refactors MainPresenter and adds unit tests

ferdyrod:


app/src/main/java/org/dhis2/usescases/main/program/ProgramFragment.kt
```diff
@@ -140,4 +148,68 @@ class ProgramFragment : FragmentGlobalAbstract(), ProgramContract.View {
     override fun clearFilters() {
         (activity as MainActivity).adapter?.notifyDataSetChanged()
     }
+
+    override fun navigateTo(program: ProgramViewModel) {
+        val bundle = Bundle()
+        val idTag = if (program.typeName() == "DataSets") {
```
Balcan: We should move this hardcoded strings as constants.

app/src/main/java/org/dhis2/usescases/main/program/ProgramPresenter.kt
```diff
@@ -1,55 +1,33 @@
...
-            FilterManager.getInstance().asFlowable()
-                .startWith(FilterManager.getInstance())
+    fun init() {
+        disposable.add(
+            filterManager.asFlowable()
+                .startWith(filterManager)
                 .doOnNext { Timber.tag("INIT DATA").d("NEW FILTER") }
                 .switchMap { filterManager ->
-                    loadingProcessor.onNext(true)
+                    view.showFilterProgress()
```
Balcan: I thought I changed this as it can cause a crash as we are accessing a ui view from an io thread

app/src/main/java/org/dhis2/usescases/main/program/ProgramPresenter.kt
```diff
@@ -72,24 +50,14 @@ class ProgramPresenter internal constructor(
...
-                    view.swapProgramModelData(),
-                    Consumer { throwable -> view.renderError(throwable.message ?: "") },
-                    Action { Timber.tag("INIT DATA").d("LOADING ENDED") }
+                    { programs -> view.swapProgramModelData(programs) },
+                    { throwable -> view.renderError(throwable.message ?: "") },
+                    { Timber.tag("INIT DATA").d("LOADING ENDED") }
                 )
         )
 
-        compositeDisposable!!.add(
```
Balcan: Check my last comment.

app/src/main/java/org/dhis2/usescases/main/program/ProgramPresenter.kt
JaimeToca: true, this is emitting in IO scheduler

ferdyrod: done

JaimeToca:
@ferdyrod I think you need to update the SDK submodule

app/src/main/java/org/dhis2/usescases/main/program/ProgramPresenter.kt
```diff
@@ -99,110 +79,42 @@ class ProgramPresenter internal constructor(
...
-        bundle.putString(idTag, programModel.id())
-        bundle.putString(Constants.DATA_SET_NAME, programModel.title())
-        bundle.putString(
-            Constants.ACCESS_DATA,
-            java.lang.Boolean.toString(programModel.accessDataWrite())
-        )
-        val programTheme = ColorUtils.getThemeFromColor(programModel.color())
-        val prefs = view!!.abstracContext.getSharedPreferences(
-            Constants.SHARE_PREFS, Context.MODE_PRIVATE
-        )
```
JaimeToca: very nice dependency decoupling. For next time, I would have injected the analytics controller and track in the presenter

app/src/main/java/org/dhis2/usescases/main/program/ProgramPresenter.kt
```diff
@@ -99,110 +79,42 @@ class ProgramPresenter internal constructor(
...
     }
 
-    override fun onSyncStatusClick(program: ProgramViewModel) {
-        view!!.showSyncDialog(
-            SyncStatusDialog.Builder()
-                .setConflictType(
-                    if (program.typeName() != "DataSets") {
-                        SyncStatusDialog.ConflictType.PROGRAM
-                    } else {
-                        SyncStatusDialog.ConflictType.DATA_SET
```
JaimeToca: For next time, SyncStatusDialog can actually be part of the program view model and use as an attribute when initialization of the dialog. If not, pass it as parameter. At the we we'd like to be the view as dumb as possible

PR #924, opened 2019-10-28
fix: Message not triggered when completing event

Balcan:
Jira issue: [ANDROAPP-2465](https://jira.dhis2.org/browse/ANDROAPP-2465)

### DIFF
--- a/app/src/main/java/org/dhis2/usescases/teiDashboard/DashboardRepositoryImpl.java
+++ b/app/src/main/java/org/dhis2/usescases/teiDashboard/DashboardRepositoryImpl.java
@@ -77,39 +77,6 @@ public class DashboardRepositoryImpl implements DashboardRepository {
             "WHERE Enrollment.trackedEntityInstance = ? AND Enrollment.program = ?\n" +
             "ORDER BY Note.storedDate DESC";
 
-    private final String PROGRAM_STAGE_FROM_EVENT = String.format(
-            "SELECT %s.* FROM %s JOIN %s " +
-                    "ON %s.%s = %s.%s " +
-                    "WHERE %s.%s = ? " +
-                    "LIMIT 1",
-            "ProgramStage", "ProgramStage", "TABLE",
-            "ProgramStage", "uid", "Event", "programStage",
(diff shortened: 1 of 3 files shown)


app/src/test/java/org/dhis2/usescases/teiDashboard/DashboardRepositoryImplTest.kt
```diff
@@ -0,0 +1,63 @@
...
+        whenever(d2.eventModule().events().uid("event_uid").get()) doReturn Single.just(getMockSingleEvent())
+
+        whenever(d2.programModule()) doReturn mock()
+        whenever(d2.programModule().programStages()) doReturn mock()
+        whenever(d2.programModule().programStages().uid("program_stage")) doReturn mock()
+        whenever(d2.programModule().programStages().uid("program_stage").get()) doReturn Single.just(getMockStage())
+
+        val testObserver = repository.displayGenerateEvent("event_uid").test()
+
+        testObserver.assertValue(getMockStage())
```
JaimeToca: I think you can assert there was no error with "assertNoErrors()". Also if Is a single may be you can also test assertValueCount.

Do you think we can verify
d2.programModule().programStages().uid(stageUid).get() from the flatmap body?

app/src/main/java/org/dhis2/usescases/teiDashboard/DashboardRepositoryImpl.java
```diff
@@ -77,39 +77,6 @@
             "WHERE Enrollment.trackedEntityInstance = ? AND Enrollment.program = ?\n" +
             "ORDER BY Note.storedDate DESC";
 
-    private final String PROGRAM_STAGE_FROM_EVENT = String.format(
```
JaimeToca: good :)

app/src/test/java/org/dhis2/usescases/teiDashboard/DashboardRepositoryImplTest.kt
Balcan: Done!

PR #925, opened 2019-10-29
fix: all fields required and mandatory fields when complete

mmmateos:
BREAKING CHANGES: Migrate DataValuePresenter to Kotlin
Check all fields required and mandatory fields when completing dataSet
Jira Issues:
https://jira.dhis2.org/browse/ANDROAPP-2470
https://jira.dhis2.org/browse/ANDROAPP-2468


app/src/main/java/org/dhis2/usescases/datasets/dataSetTable/dataSetSection/DataValuePresenter.kt
```diff
@@ -0,0 +1,741 @@
...
+                            table.indexOf(row),
+                            true
+                        )
+                    }
+                }
+            }
+        }
+        return mandatoryOk
+    }
+
```
mmmateos: These are the modified methods

app/src/main/java/org/dhis2/usescases/datasets/dataSetTable/dataSetSection/DataValuePresenter.kt
```diff
@@ -0,0 +1,741 @@
...
+                            List<DataSetTableModel>,
+                            List<DataElementOperand>,
+                            List<DataElementOperand>,
+                            Sextet<CategoryCombo,
+                                List<DataElement>,
+                                Map<String, List<List<Pair<CategoryOption, Category>>>>,
+                                List<DataSetTableModel>,
+                                List<DataElementOperand>,
+                                List<DataElementOperand>>>
+                        { val0, val1, val2, val3, val4, val5 ->
```
Balcan: val0, val1...val5 I don't know what are these variables.
Not for now but we will have to check this together :)

app/src/main/java/org/dhis2/usescases/datasets/dataSetTable/dataSetSection/DataValuePresenter.kt
```diff
@@ -0,0 +1,741 @@
...
+                    )
+
+                    setTableData(dataTableModel!!)
+                }
+                .subscribeOn(schedulerProvider.io())
+                .observeOn(schedulerProvider.ui())
+                .subscribe(
+                    { quartet ->
+                        view.setTableData(
+                            quartet.val0(),
```
Balcan: in kotlin you can add a variable name so you could do something like:
variableName1 = quartet.val0(),
variableName2 = quartet.val1(),
...

app/src/main/java/org/dhis2/usescases/datasets/dataSetTable/dataSetSection/DataValuePresenter.kt
```diff
@@ -0,0 +1,741 @@
...
+        val isEditable = accessDataWrite &&
+            !isExpired(dataSet) &&
+            dataInputPeriodModel.isEmpty() || (
+            checkHasInputPeriod() != null && DateUtils.getInstance().isInsideInputPeriod(
+                checkHasInputPeriod()
+            )
+            ) &&
+            !isApproval
+
+        return Quartet.create(dataTableModel, listFields, cells, isEditable)
```
Balcan: Not for now. But this method is way too long (155 lines).
Would be better to split it into several smaller methods.

app/src/main/java/org/dhis2/usescases/datasets/dataSetTable/dataSetSection/DataValuePresenter.kt
Balcan: Just so the others reviewers know:

- checkAllFieldRequired
- checkMandatoryField

app/src/main/java/org/dhis2/usescases/datasets/dataSetTable/dataSetSection/DataValuePresenter.kt
ferdyrod: Also, in kotlin we can deconstruct an object
    ` val (varName1, varName2, varNmae3, varName4) = quartet`

and varName1 will have quartet.val0() value and so forth.

app/src/main/java/org/dhis2/usescases/datasets/dataSetTable/dataSetSection/DataValuePresenter.kt
JaimeToca: any unit test?

app/src/main/java/org/dhis2/usescases/datasets/dataSetTable/dataSetSection/DataValuePresenter.kt
JaimeToca: or classes I see too many responsabilities in same method and things that perhaps the presenter should not do. Let's take into account next time when performing unit tests after the release.

PR #926, opened 2019-10-29
fix: render period dialog when there are open future periods

mmmateos:
Jira Issue: https://jira.dhis2.org/browse/ANDROAPP-2469


app/src/main/java/org/dhis2/utils/customviews/PeriodAdapter.kt
```diff
@@ -0,0 +1,80 @@
...
+
+    companion object {
+        const val DEFAULT_PERIODS_SIZE = 10
+    }
+
+    private val datePeriods: MutableList<Date>
+    private var onDateSetListener: PeriodDialog.OnDateSet? = null
+    private var lastDate: Date
+
+    init {
```
Balcan: You could inline this init with the variables

ferdyrod: I believe that the way the it's done is the right way. According to kotlin's documentation _since the primary constructor cannot contain any code. Initialization code can be placed in initializer block_. but the declaration of the variables doesn't have to be inside.

app/src/main/java/org/dhis2/utils/customviews/PeriodAdapter.kt
```diff
@@ -0,0 +1,80 @@
...
+    private fun setDates() {
+        repeat(DEFAULT_PERIODS_SIZE) {
+            datePeriods.add(lastDate)
+            lastDate = DateUtils.getInstance().getNextPeriod(periodType, lastDate, -1)
+        }
+    }
+
+    override fun getItemCount(): Int {
+        return datePeriods.size + 1
+    }
```
ferdyrod: Since this method return a single expression. We can omit the braces, the return keyword and the return type, if the compiler can infer its type. so we can do a one-liner method.

`override fun getItemCount() = datePeriods.size + 1`

app/src/main/java/org/dhis2/utils/customviews/PeriodAdapter.kt
```diff
@@ -0,0 +1,80 @@
...
+
+    private val datePeriods: MutableList<Date>
+    private var onDateSetListener: PeriodDialog.OnDateSet? = null
+    private var lastDate: Date
+
+    init {
+        datePeriods = ArrayList()
+        lastDate = DateUtils.getInstance()
+            .getNextPeriod(periodType, DateUtils.getInstance().today, openFuturePeriods - 1)
+        setDates()
```
JaimeToca: I see the iniitialization of the old class as :  this.openFuturePeriods = openFuturePeriods;
I have some questions, why are we doing openFuturePeriods - 1 ? Can openFuturePeriods be 0 ?

app/src/main/java/org/dhis2/utils/customviews/PeriodAdapter.kt
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    override fun getItemCount(): Int {
+        return datePeriods.size + 1
+    }
+
+    fun setOnDateSetListener(onDateSetListener: PeriodDialog.OnDateSet) {
+        this.onDateSetListener = onDateSetListener
+    }
+}
```
JaimeToca: I think we can use a kotlin lambda and pass it as the last parameter In the constructor: 

_val listener: (PeriodDialog.OnDateSet) -> Unit_ 

Then above in the "onClick" you will do this _listener(datePeriods[holder.adapterPosition])_

and outside this class you will have something like

MyAdapter(arguments){
            //do something with the listener
        }

app/src/main/java/org/dhis2/utils/customviews/PeriodAdapter.kt
mmmateos: Nice, thank you

PR #928, opened 2019-10-29
fix: Images were not uploading to server

ferdyrod:
This pull requests refers to jira issue [ANDROAPP-2463](https://jira.dhis2.org/browse/ANDROAPP-2463).

Besides fixing the issue it converts some classes to Kotlin and prepares the sync presenter to testing, as well as does dependency injection on the sync presenter.


app/src/main/java/org/dhis2/data/service/ServiceModule.java
```diff
@@ -23,7 +25,13 @@ NotificationManager notificationManager(@NonNull Context context) {
...
     @Provides
     @PerService
-    SyncPresenter syncPresenter(@NonNull D2 d2) {
-        return new SyncPresenterImpl(d2);
+    SyncPresenter syncPresenter(
+            @NonNull Context context,
+            @NonNull D2 d2,
+            @NonNull PreferenceProvider preferences
+    ) {
+        WorkManager workManager = WorkManager.getInstance(context);
```
Balcan: We could add the WorkManager in the application module so that we can inject it.

app/src/main/java/org/dhis2/data/service/SyncGranularWorker.kt
```diff
@@ -0,0 +1,82 @@
...
+) : Worker(context, workerParams) {
+
+    @Inject
+    internal lateinit var presenter: SyncPresenter
+
+    override fun doWork(): Result {
+        Objects.requireNonNull((applicationContext as App).userComponent())!!
+            .plus(SyncGranularRxModule()).inject(this)
+
+        val uid = inputData.getString(UID)?.let { it } ?: return Result.failure()
```
JaimeToca: you can also use the elvis operator like : inputData.?getString(UID) ?: Result.failure()

app/src/main/java/org/dhis2/data/service/SyncPresenterImpl.kt
```diff
@@ -0,0 +1,368 @@
...
+import org.dhis2.data.prefs.PreferenceProvider
+import org.dhis2.utils.DateUtils
+import org.hisp.dhis.android.core.D2
+import org.hisp.dhis.android.core.arch.call.D2Progress
+import org.hisp.dhis.android.core.common.State
+import org.hisp.dhis.android.core.imports.TrackerImportConflict
+import org.hisp.dhis.android.core.program.ProgramType
+import timber.log.Timber
+
+class SyncPresenterImpl(
```
JaimeToca: For next time lets try to have this kind of migration outside testing cycles fixes pls

app/src/main/java/org/dhis2/data/service/ServiceModule.java
ferdyrod: You are right, the WorkManager is already added in the appModule. There is no need to do this anymore. Will remove

app/src/main/java/org/dhis2/data/service/SyncGranularWorker.kt
ferdyrod: Can't do that here, because the variable **uid** is later used so can't have it be a Result.failure().
What I can do is remove the `let { it }` whichis not needed and that will work.

`val uid = inputData.getString(UID) ?: return Result.failure()`

app/src/main/java/org/dhis2/data/service/SyncPresenterImpl.kt
ferdyrod: noted 👌

PR #937, opened 2019-11-04
test: checkAllFieldsRequired and checkMandatoryFields

mmmateos:


app/src/main/java/org/dhis2/usescases/datasets/dataSetTable/dataSetSection/DataValuePresenter.kt
```diff
@@ -560,11 +562,15 @@ class DataValuePresenter(
         }
     }
 
-    private fun checkAllFieldRequired(): Boolean {
+    @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
```
ferdyrod: Nice!
💯

PR #939, opened 2019-11-04
refactor: removed deprecated sdk

mmmateos:


app/src/main/java/org/dhis2/utils/DateUtils.java
```diff
@@ -1153,7 +1155,8 @@ public Boolean isEventExpired(Date eventDate, Date completeDate, EventStatus sta
                 expDate = calendar.getTime();
             }
 
-            expiredBecouseOfPeriod = expDate != null && expDate.before(getCalendar().getTime());
+//            expiredBecouseOfPeriod = expDate != null && expDate.before(getCalendar().getTime());
```
JaimeToca: remove it?

app/src/test/java/org/dhis2/utils/DateUtilsTest.java
```diff
@@ -699,14 +697,15 @@ public void active_event_NcD_Monthly_0_is_expired() throws ParseException {
     }
 
     @Test
+    @Ignore
```
JaimeToca: does it makes sense to remove it?

PR #944, opened 2019-11-06
fix(home): Home takes too long to load for dataSets

Balcan:
Jira issue: [ANDROAPP-2480]https://jira.dhis2.org/browse/ANDROAPP-2480)

### DIFF
--- a/app/src/main/java/org/dhis2/usescases/main/program/HomeRepositoryImpl.kt
+++ b/app/src/main/java/org/dhis2/usescases/main/program/HomeRepositoryImpl.kt
@@ -57,60 +57,21 @@ internal class HomeRepositoryImpl(private val d2: D2, private val eventLabel: St
                     count = repo.blockingCount()
                 }
 
-                var state: State? = State.SYNCED
-
-                dataSet.dataSetElements()!!.forEach {
-                    val dataElement =
-                        d2.dataElementModule().dataElements().uid(it.dataElement().uid())
-                            .blockingGet()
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/dhis2/usescases/main/program/HomeRepositoryImpl.kt
```diff
@@ -57,60 +57,21 @@ internal class HomeRepositoryImpl(private val d2: D2, private val eventLabel: St
...
+                        possibleStates.contains(State.WARNING)){
+                    State.WARNING
+                }else if(possibleStates.contains(State.SENT_VIA_SMS) ||
+                        possibleStates.contains(State.SYNCED_VIA_SMS)){
+                    State.SENT_VIA_SMS
+                }else if(possibleStates.contains(State.TO_UPDATE) ||
+                        possibleStates.contains(State.TO_POST)){
+                    State.TO_UPDATE
+                }else{
+                    State.SYNCED
```
JaimeToca: very nice, u could use "when" kotlin if u want to make it more beautiful :D

ferdyrod: if we have time, we should pass the ktlint task.

Other than that, great!

Balcan: Last PR in the queue has the klint check :)

PR #947, opened 2019-11-06
fix(tableView): scroll to start of table when zoom is pressed

mmmateos:
https://jira.dhis2.org/browse/ANDROAPP-2481


Balcan:
Tested!

PR #949, opened 2019-11-06
fix: (SyncStatusDialog): Adds onDismissListener to dialog

ferdyrod:
Related to PR #945 

Adds ondismiss listener to dialog to prevent crash. Also changes the name of the method to show the sync dialog.


app/src/main/java/org/dhis2/usescases/datasets/dataSetTable/DataSetTablePresenter.java
```diff
@@ -140,4 +140,16 @@ public String getCatOptComboFromOptionList(List<String> catOpts) {
         return tableRepository.getCatOptComboFromOptionList(catOpts);
     }
 
+    @Override
+    public void updateState(){
```
JaimeToca: unit test ?

ferdyrod: Added!

PR #952, opened 2019-11-06
fix: DataSet state not updated with complete/reopen actions

Balcan:
[jira issue](https://jira.dhis2.org/browse/ANDROAPP-2487)


mmmateos:
Tested :ok_hand:

PR #958, opened 2019-11-08
test(program rules): RulesUtilsProvider unit tests

Balcan:
- Added unit tests to RulesUtilsProvider
- Added TextUtils to test\java\android\text

[ANDROAPP-2498](https://jira.dhis2.org/browse/ANDROAPP-2498)


app/src/main/java/org/dhis2/utils/RulesUtilsProviderImpl.kt
```diff
@@ -145,19 +137,21 @@ class RulesUtilsProviderImpl(private val codeGenerator: CodeGenerator) : RulesUt
...
         fieldViewModels: MutableMap<String, FieldViewModel>,
-        rulesActionCallbacks: RulesActionCallbacks
+        rulesActionCallbacks: RulesActionCallbacks,
+        data: String
     ) {
         val model = fieldViewModels[showError.field()]
 
         if (model != null) {
-            fieldViewModels[showError.field()] = model.withError(showError.content())
+            fieldViewModels[showError.field()] = model.withError(showError.content() + " " + data)
```
ferdyrod: With kotlin we can use String interpolation to concatenate strings

`fieldViewModels[showError.field()] = model.withError("${showError.content()} $data")`

when you have to use a method or property of the variable it needs to be wrapped with the { } signs.

app/src/test/java/org/dhis2/utils/RulesUtilsProviderImplTest.kt
```diff
@@ -0,0 +1,452 @@
...
+            actionCallbacks
+        )
+
+        verify(actionCallbacks, times(1)).setHideProgramStage(testingUid)
+    }
+
+    @Test
+    fun `RuleActionHideProgramStage should remove stage from possible selections`() {
+
+    }
```
JaimeToca: empty? :D

Balcan: :O

app/src/main/java/org/dhis2/utils/RulesUtilsProviderImpl.kt
Balcan: Done

PR #959, opened 2019-11-08
test(DataSetTable): [ANDROAPP-2496] Adds missing test to DataSetTablePresenter

ferdyrod:
**JIRA ISSUE** [ANDROAPP-2496](https://jira.dhis2.org/browse/ANDROAPP-2496)


app/src/main/java/org/dhis2/usescases/datasets/dataSetTable/DataSetTablePresenter.kt
```diff
@@ -0,0 +1,109 @@
...
+    val disposable = CompositeDisposable()
+
+    fun init(catCombo: String) {
+        disposable.add(
+            tableRepository.sections
+                .subscribeOn(schedulerProvider.io())
+                .observeOn(schedulerProvider.ui())
+                .subscribe(
+                    { view.setSections(it) },
+                    { Timber.e(it) }
```
MarinaCape: it can be simplify putting view::setSections and Timber::e Without { }

PR #961, opened 2019-11-11
fix: enable orgUnit field when event is referral

mmmateos:
Jira Issue: https://jira.dhis2.org/browse/ANDROAPP-2500


app/src/main/java/org/dhis2/usescases/eventsWithoutRegistration/eventInitial/EventInitialActivity.java
```diff
@@ -409,7 +409,7 @@ public void setProgram(@NonNull Program program) {
             binding.date.setText(selectedDateString);
             if(selectedOrgUnit == null)
                 presenter.initOrgunit(selectedDate);
-            else
+            else if (eventCreationType != EventCreationType.REFERAL)
                 binding.orgUnit.setEnabled(false);
```
ferdyrod: Should we do something else if neither of these conditions are met?

mmmateos: I don't think so, in that case we don't have to do anything with the orgUnit.

ferdyrod:
This is already tested on device!

PR #962, opened 2019-11-11
fix(Search): Fixes search teis list when device is offline

ferdyrod:
**Jira Issue**: [ANDROAPP-2502](https://jira.dhis2.org/browse/ANDROAPP-2502)


mmmateos:
Tested :ok_hand:

PR #963, opened 2019-11-11
test: adds test on DataSetInitial

mmmateos:
BREAKING CHANGE: Migrates DataSetInitial to kotlin and adds test
https://jira.dhis2.org/browse/ANDROAPP-2494


app/src/test/java/org/dhis2/usecases/datasets/dataSetInitial/DataSetInitialPresenterTest.kt
```diff
@@ -0,0 +1,141 @@
...
+    }
+
+    @Test
+    fun `Should navigate to dataSetTableActivity when actionbutton is clicked`() {
+        val catCombo = "catComboUid"
+        val catOptionCombo = "catOptionCombo"
+        val periodId = "periodId"
+        val periodType = PeriodType.Monthly
+
+        `Should setOrgUnit and data`()
```
JaimeToca: you don't need to execute the other unit test. You can just call the whenevers :+1:

app/src/test/java/org/dhis2/usecases/datasets/dataSetInitial/DataSetInitialPresenterTest.kt
```diff
@@ -0,0 +1,141 @@
...
+            PeriodType.Daily,
+            mutableListOf(),
+            0
+            )
+        whenever(repository.orgUnits()) doReturn Observable.just(orgUnits)
+        whenever(repository.dataSet()) doReturn Observable.just(dataSet)
+
+        presenter.init()
+
+        verify(view).setOrgUnit(orgUnits[0])
```
JaimeToca: I think we're missing the case when orgUnits size == 0 or > 1

app/src/main/java/org/dhis2/usescases/datasets/datasetInitial/DataSetInitialPresenter.kt
```diff
@@ -0,0 +1,124 @@
...
+                .subscribeOn(schedulerProvider.io())
+                .observeOn(schedulerProvider.ui())
+                .subscribe(
+                    { orgUnits ->
+                        this.orgUnits = orgUnits
+                        if (orgUnits.size == 1) {
+                            view.setOrgUnit(orgUnits[0])
+                        }
+                    },
+                    { Timber.d(it) }
```
MarinaCape: It can be simplify with Timber::it and without { }

app/src/test/java/org/dhis2/usecases/datasets/dataSetInitial/DataSetInitialPresenterTest.kt
JaimeToca: just saw the test below :D

Balcan:
@mmmateos You can merge it now :) Good job!

PR #965, opened 2019-11-12
fix(enrollment): Missing mandatory fields not showing error in TEI details

Balcan:
[jira issue](https://jira.dhis2.org/browse/ANDROAPP-2506)


Balcan:
Before reviewing this issue wait for the previous ones to be merged or just check the EnrollmentActivity.kt file where the changes related to this PR has been made.

app/src/main/java/org/dhis2/usescases/enrollment/EnrollmentActivity.kt
```diff
@@ -247,7 +247,19 @@ class EnrollmentActivity : ActivityGlobalAbstract(), EnrollmentContract.View {
     }
 
     override fun goBack() {
-        onBackPressed()
+        if (!adapter.mandatoryOk()) {
```
JaimeToca: can you move it to presenter and expose two methods from the view, so doing this we can unit test, cover this bug and forget forever :D

ferdyrod:
This is already tested on device!

PR #968, opened 2019-11-12
test(DataSetDetail): Adds unit test to presenter, migrates to kotlin

ferdyrod:
**JIRA ISSUE**: [ANDROAPP-2504](https://jira.dhis2.org/browse/ANDROAPP-2504)


app/src/main/java/org/dhis2/usescases/datasets/datasetDetail/DataSetDetailActivity.kt
```diff
@@ -0,0 +1,242 @@
...
+
+    override fun startNewDataSet() {
+        binding.addDatasetButton.isEnabled = false
+        val bundle = Bundle()
+        bundle.putString(Constants.DATA_SET_UID, dataSetUid)
+        startActivity(DataSetInitialActivity::class.java, bundle, false, false, null)
+    }
+
+    override fun openDataSet(dataSet: DataSetDetailModel) {
+        val bundle = Bundle()
```
MarinaCape: Could use 'with' method or 'run' extension function to avoid use all times bundle word

PR #969, opened 2019-11-13
fix: Physical back button not showing empty/error field dialog

Balcan:
From the TEI details screen in the dashboard. Clicking the device physical back button was not triggering data verification and because of it you could leave the screen with empty values in mandatory fields.


app/src/main/java/org/dhis2/usescases/enrollment/EnrollmentActivity.kt
```diff
@@ -233,69 +238,73 @@ class EnrollmentActivity : ActivityGlobalAbstract(), EnrollmentView {
...
 
     override fun onBackPressed() {
-        if (mode == EnrollmentMode.CHECK) super.onBackPressed() else showDeleteDialog()
+        if (mode == EnrollmentMode.CHECK) {
+            if (presenter.dataIntegrityCheck(adapter.mandatoryOk(), adapter.hasError())) {
+                super.onBackPressed()
+            }
+        } else {
+            showDeleteDialog()
+        }
```
JaimeToca: To me is okay :+1: . If you want, I would have done something.
`override fun onBackPressed() {
  presenter.onBackPressed(EnrollmentMode.Check)
}`

And then move as much logic to presenter, then unit test it and forget about this bug forever :)

mmmateos:
Tested :ok_hand:

PR #970, opened 2019-11-13
fix: Missing Tracked Entity Type icon in dashboard

Balcan:
[jira issue](https://jira.dhis2.org/browse/ANDROAPP-2509)


mmmateos:
Tested :ok_hand:

PR #971, opened 2019-11-13
fix(FilterManager): fixes bug with the period filter from-To and other

ferdyrod:
**JIRA ISSUE**: [ANDROAPP-2512](https://jira.dhis2.org/browse/ANDROAPP-2512)

### DIFF
--- a/app/src/main/java/org/dhis2/utils/filters/PeriodFilterHolder.java
+++ b/app/src/main/java/org/dhis2/utils/filters/PeriodFilterHolder.java
@@ -65,8 +65,10 @@ class PeriodFilterHolder extends FilterHolder implements CompoundButton.OnChecke
                 break;
             case R.id.fromTo:
                 localBinding.periodLayout.fromTo.setChecked(true);
+                localBinding.periodLayout.other.setChecked(false);
                 break;
             case R.id.other:
+                localBinding.periodLayout.fromTo.setChecked(false);
                 localBinding.periodLayout.other.setChecked(true);
                 break;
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/dhis2/utils/filters/PeriodFilterHolder.java
```diff
@@ -84,18 +84,8 @@ private void setListeners(FilterPeriodBinding periodLayout) {
...
-            }
-        });
-        periodLayout.other.setOnClickListener(view -> {
-            if (periodLayout.other.isChecked()) {
-                updateSelection(R.id.other);
-                FilterManager.getInstance().addPeriodRequest(FilterManager.PeriodRequest.OTHER);
-            }
-        });
+        periodLayout.fromTo.setOnCheckedChangeListener(this);
+        periodLayout.other.setOnCheckedChangeListener(this);
```
Balcan: This was using the onclicklistener to be able to select again these filters even if they are already selected.
Have you test to click them once they are checked?

ferdyrod: Ok, will have to change it. Didn't know that that was a possibility.

Balcan: Yep, it was discussed in this [issue](https://jira.dhis2.org/browse/ANDROAPP-2401)

Balcan:
Tested!

PR #975, opened 2019-11-14
fix(TrackedEntityAttributeValue): Removes uid from Search and Dashboard

ferdyrod:
**JIRA ISSUE**: [ANDROAPP-2516](https://jira.dhis2.org/browse/ANDROAPP-2516)

### DIFF
--- a/app/src/main/java/org/dhis2/usescases/teiDashboard/DashboardRepositoryImpl.java
+++ b/app/src/main/java/org/dhis2/usescases/teiDashboard/DashboardRepositoryImpl.java
@@ -54,7 +54,6 @@ import android.content.ContentValues;
 import android.content.Context;
 import android.content.res.Resources;
 import android.database.Cursor;
-import android.database.sqlite.SQLiteStatement;
 
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/dhis2/usescases/teiDashboard/DashboardRepositoryImpl.java
```diff
@@ -374,12 +373,34 @@ public void setDefaultCatOptCombToEvent( String eventUid )
...
+                    .map(values -> {
+                        List<TrackedEntityAttributeValue> toRemove = new ArrayList<>();
+                        for (TrackedEntityAttributeValue value: values) {
+                            ValueType type = d2.trackedEntityModule().trackedEntityAttributes().uid(value.trackedEntityAttribute()).blockingGet().valueType();
+                            if (type == ValueType.IMAGE) {
+                                toRemove.add(value);
+                            }
+                        }
+                        values.removeAll(toRemove);
+                        return values;
```
JaimeToca: Do we need to perform this in the IO?

app/src/main/java/org/dhis2/usescases/teiDashboard/DashboardRepositoryImpl.java
```diff
@@ -374,12 +373,34 @@ public void setDefaultCatOptCombToEvent( String eventUid )
     {
         if ( programUid != null )
             return briteDatabase
-                .createQuery( ATTRIBUTE_VALUES_TABLE, ATTRIBUTE_VALUES_QUERY, programUid, teiUid == null ? "" : teiUid )
-                .mapToList( cursor -> ValueUtils.transform( briteDatabase, cursor ) );
+                    .createQuery( ATTRIBUTE_VALUES_TABLE, ATTRIBUTE_VALUES_QUERY, programUid, teiUid == null ? "" : teiUid )
```
MarinaCape: Could it change the query to modules from D2 ?

app/src/main/java/org/dhis2/usescases/teiDashboard/DashboardRepositoryImpl.java
ferdyrod: I believe the IO is the perfect place to do database operations within the repository.

JaimeToca: yes I agree !!! my question was the I did not see scheduler.io in any place, is this being done inside the sdk?

app/src/main/java/org/dhis2/usescases/teiDashboard/DashboardRepositoryImpl.java
ferdyrod: For the time being, we decided not to change it to the sdk modules. This will be done in the near future.

Balcan:
I think @JaimeToca gave his approval in the daily. Don't you?

PR #978, opened 2019-11-15
fix: Crash when selecting Report Jira Issue in home screen

Balcan:
[jira issue](https://jira.dhis2.org/browse/ANDROAPP-2519)

### DIFF
--- a/app/src/main/java/org/dhis2/usescases/jira/JiraViewModel.kt
+++ b/app/src/main/java/org/dhis2/usescases/jira/JiraViewModel.kt
@@ -12,10 +12,8 @@ import okhttp3.MediaType
 import okhttp3.OkHttpClient
 import okhttp3.RequestBody
 import okhttp3.ResponseBody
-import org.dhis2.Bindings.default
 import org.dhis2.data.prefs.PreferenceProvider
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/dhis2/usescases/jira/JiraViewModel.kt
```diff
@@ -45,6 +57,10 @@ class JiraViewModel : ViewModel(), JiraActions {
         issueService = retrofit.create<JiraIssueService>(JiraIssueService::class.java)
         prefs = preferenceProvider
```
ferdyrod: This is a little redundant.

JaimeToca:
@Balcan do we need to populate the jira sharepreferences for newer versions in the splash?

Balcan:
> @Balcan do we need to populate the jira sharepreferences for newer versions in the splash?

Nope we don't need it :)
It doesn't have such an impact. They will just need to enter their jira credentials again.

app/src/main/java/org/dhis2/usescases/jira/JiraViewModel.kt
Balcan: true

PR #983, opened 2019-11-18
test(SyncPresenter): Adds unit tests to SyncPresenter. creates WorkManagerController

ferdyrod:
**JIRA ISSUE**: [ANDROAPP-2524](https://jira.dhis2.org/browse/ANDROAPP-2524)


ferdyrod:
The WorkManagerController is very simple and it needs more methods and maybe a better way to call the methods that are already included.

app/src/main/java/org/dhis2/usescases/sync/SyncActivity.kt
```diff
@@ -0,0 +1,201 @@
...
+            WorkInfo.State.SUCCEEDED -> {
+                binding.metadataText.text = getString(R.string.configuration_ready)
+                Bindings.setDrawableEnd(
+                    binding.metadataText,
+                    AppCompatResources.getDrawable(this, R.drawable.animator_done)
+                )
+                presenter.getTheme()
+            }
+            else -> {
+            }
```
Balcan: We should add a comment on why the else branch is empty

app/src/main/java/org/dhis2/usescases/sync/SyncActivity.kt
```diff
@@ -0,0 +1,201 @@
...
+                binding.eventsText.text = getString(R.string.data_ready)
+                Bindings.setDrawableEnd(
+                    binding.eventsText,
+                    AppCompatResources.getDrawable(this, R.drawable.animator_done)
+                )
+                presenter.syncReservedValues()
+                startMain()
+            }
+            else -> {
+            }
```
Balcan: We should add a comment on why the else branch is empty

JaimeToca:
hey ferdy, is it pointing to the right rule engine?

ferdyrod:
> hey ferdy, is it pointing to the right rule engine?

@JaimeToca Yes, the rule-engine master branch was updated so this adds that link to the latest commit

JaimeToca:
Great @ferdyrod, this PR looks great to me.  I'd test it the sync in this jira issue

PR #1007, opened 2019-11-22
test: Unit test in IndicatorsPresenter and migrate to kotlin

MarinaCape:
https://jira.dhis2.org/browse/ANDROAPP-2528


app/src/main/java/org/dhis2/usescases/teiDashboard/dashboardfragments/indicators/IndicatorsPresenterImpl.java
```diff
@@ -91,9 +88,9 @@ public void init(IndicatorsContracts.View view) {
                                 }))
                         .subscribeOn(schedulerProvider.io())
                         .observeOn(schedulerProvider.ui())
-                        .subscribe(
-                                view.swapIndicators(),
-                                Timber::d
+                        .subscribe(data -> {view.swapIndicators(data);},
+                                //view.swapIndicators(),
```
ferdyrod: We should not leave commented code, unless we are going to use it later.

PR #1025, opened 2019-11-25
fix: do not overlap label by orgUnit name

mmmateos:
Jira Issue: https://jira.dhis2.org/browse/ANDROAPP-2538


ferdyrod:
Tested on various devices! 👍

PR #1026, opened 2019-11-26
fix(programRules): Event forms takes too long to init

Balcan:
[jira issue](https://jira.dhis2.org/browse/ANDROAPP-2539)


app/src/main/java/org/dhis2/usescases/eventsWithoutRegistration/eventCapture/EventCaptureFragment/EventCaptureFormFragment.java
```diff
@@ -99,11 +100,19 @@ public View onCreateView(@NonNull LayoutInflater inflater, @Nullable ViewGroup c
...
     }
 
+    @Override
+    public void onResume() {
+        super.onResume();
+        new Handler().post(() -> {
+            activity.getPresenter().init();
+            activity.getPresenter().initCompletionPercentage(sectionSelectorAdapter.completionPercentage());
+        });
+    }
```
JaimeToca: why use a handler?

app/src/main/java/org/dhis2/usescases/eventsWithoutRegistration/eventCapture/EventCaptureActivity.java
```diff
@@ -91,14 +91,17 @@ protected void onCreate(@Nullable Bundle savedInstanceState) {
...
+
+//        presenter.init(this);
 
     }
 
 
     @Override
     protected void onResume() {
         super.onResume();
+        setUp();
```
JaimeToca: is this method empy?

Balcan: Removed

PR #1028, opened 2019-11-26
fix(QR): Wrong sharing actions for events and TEI

Balcan:
[jira issue](https://jira.dhis2.org/browse/ANDROAPP-2544)


app/src/main/java/org/dhis2/usescases/teiDashboard/dashboardfragments/tei_data/TEIDataPresenterImpl.java
```diff
@@ -225,43 +216,10 @@ public void onFollowUp(DashboardProgramModel dashboardProgramModel) {
...
-                default:
-                    return true;
-            }
-        });
-
-        menu.show();
+        view.analyticsHelper().setEvent(TYPE_SHARE, TYPE_QR, SHARE_TEI);
+        Intent intent = new Intent(view.getContext(), QrActivity.class);
+        intent.putExtra("TEI_UID", teiUid);
+        view.showQR(intent);
```
ferdyrod: This should be on the view, as well as, using the AnalyticsHelper instead of the deprecated view.analyticsHelper()

Balcan: True. I'll fix it

PR #1034, opened 2019-11-27
fix(Forms): Eliminates the need to click twice to get the cursor on the chosen field

ferdyrod:
**JIRA ISSUE**: [ANDROAPP-2415](https://jira.dhis2.org/browse/ANDROAPP-2415)


app/src/main/java/org/dhis2/data/forms/dataentry/fields/FormViewHolder.java
```diff
@@ -82,7 +82,7 @@ public void closeKeyboard(View v) {
 
     public void openKeyboard(View v) {
         InputMethodManager imm = (InputMethodManager) v.getContext().getSystemService(Activity.INPUT_METHOD_SERVICE);
-        boolean shown = imm.showSoftInput(v.findFocus(),InputMethodManager.SHOW_FORCED);
+        v.postDelayed(() -> imm.showSoftInput(v , InputMethodManager.SHOW_FORCED), 200);
        /* if (!imm.showSoftInput(v, 0))
```
Balcan: A comment... Let's remove it! :)

app/src/main/java/org/dhis2/data/forms/dataentry/DataEntryAdapter.java
```diff
@@ -261,27 +251,14 @@ public void swap(@NonNull List<FieldViewModel> updates) {
 
         diffResult.dispatchUpdatesTo(this);
 
-        if (nextFocusPosition != -1 && currentFocusPosition == lastFocusPosition && nextFocusPosition < viewModels.size())
+        if (nextFocusPosition != -1 && currentFocusPosition == lastFocusPosition && nextFocusPosition < viewModels.size()) {
             currentFocusUid.setValue(viewModels.get(nextFocusPosition).uid());
-        else if (currentFocusPosition != -1 && currentFocusPosition < viewModels.size())
+        } else if (nextFocusPosition == viewModels.size()) {
+            currentFocusUid.setValue("");
```
JaimeToca: is it needed?

ferdyrod: Yes, because when the focus is on the last field and the currentFocusUid is not updated, the last will request focus and continue to show the keyboard to the user. Basically this is done to close the keyboard on the last text field of the form.

PR #1037, opened 2019-11-27
fix: handle keypad 'next' behavior on dataSet tables

mmmateos:
Jira Issue: https://jira.dhis2.org/browse/ANDROAPP-2541

### DIFF
--- a/tableview/src/main/java/com/evrencoskun/tableview/TableView.java
+++ b/tableview/src/main/java/com/evrencoskun/tableview/TableView.java
@@ -902,13 +902,4 @@ public class TableView extends FrameLayout implements ITableView {
     public View getCornerView() {
         return getAdapter().getCornerView();
     }
-
-
-    public void selectNext() {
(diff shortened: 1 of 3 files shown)


tableview/src/main/java/com/evrencoskun/tableview/TableView.java
```diff
@@ -902,4 +902,13 @@ protected void onRestoreInstanceState(Parcelable state) {
     public View getCornerView() {
         return getAdapter().getCornerView();
     }
+
+
+    public void selectNext() {
+        if(mColumnHeaderRecyclerViews.get(mColumnHeaderRecyclerViews.size()-1).getAdapter().getItemCount() > getSelectedColumn()+ 1) {
```
JaimeToca: mColumnHeaderRecyclerViews.size()

mmmateos: It shouldn't. In that case there wouldn' be a table

PR #1038, opened 2019-11-28
fix(developActivity): Icons tested individually to catch more errors

Balcan:
[jira issue](https://jira.dhis2.org/browse/ANDROAPP-2324)


app/src/main/java/org/dhis2/usescases/development/DevelopmentActivity.java
```diff
@@ -82,21 +82,64 @@ private void renderIconForPosition(int position) {
...
+        binding.iconImageNegativeTint.setImageDrawable(null);
+
         boolean hasError = false;
         try {
             binding.iconImagePossitive.setImageResource(iconResource_positive);
+
+        } catch (Exception e) {
+            e.printStackTrace();
+            hasError = true;
+        }
```
JaimeToca: It's better to have a commond method. FetchOrCatchException() {
.........
}

PR #1041, opened 2019-11-28
fix(orgUnits): Show capture orgUnits when search ou has not been assigned in web

Balcan:
[jira issue](https://jira.dhis2.org/browse/ANDROAPP-2564)


JaimeToca:
Everything okay @Balcan  this logic should move to presentation sometime soon

PR #1045, opened 2019-11-29
fix(EnrollmentActivity): Field TEI Detail is not being saved when save without change the focus

MarinaCape:
https://jira.dhis2.org/browse/ANDROAPP-2565


mmmateos:
Tested :ok_hand:

PR #1047, opened 2019-12-02
fix(sms): Show sms permission when enabling option in settings

Balcan:
[jira issue](https://jira.dhis2.org/browse/ANDROAPP-2590)

### DIFF
--- a/app/src/main/java/org/dhis2/usescases/settings/SyncManagerFragment.java
+++ b/app/src/main/java/org/dhis2/usescases/settings/SyncManagerFragment.java
@@ -1,10 +1,12 @@
 package org.dhis2.usescases.settings;
 
 
+import android.Manifest;
 import android.app.NotificationChannel;
 import android.app.NotificationManager;
 import android.content.Context;
 import android.content.SharedPreferences;
+import android.content.pm.PackageManager;
 import android.os.Build;
 import android.os.Bundle;
 import android.os.Handler;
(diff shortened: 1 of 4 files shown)


app/src/main/java/org/dhis2/usescases/settings/SyncManagerFragment.java
```diff
@@ -247,6 +215,60 @@ public void showSmsSettings(boolean enabled, String number, boolean waitForRespo
...
+
+        listenerDisposable.add(RxCompoundButton.checkedChanges(binding.settingsSms.findViewById(R.id.settings_sms_switch))
+                .debounce(1000, TimeUnit.MILLISECONDS, Schedulers.io())
+                .observeOn(AndroidSchedulers.mainThread())
+                .subscribe(
+                        isChecked ->
+                                presenter.smsSwitch(
+                                        NetworkUtils.isOnline(context) &&
+                                                isChecked &&
+                                                checkSMSPermissions()
```
JaimeToca: shouldn't this be !NetworkUtils.isOnline(context) ?

Balcan: No, you need to be online to enable sms as it will download some extra data

JaimeToca:
everything fine @Balcan , please add info to the jira ticket regarding testing. Thanks

PR #1052, opened 2019-12-03
fix(teiDashboard): Fixed event order

Balcan:
[jira issue](https://jira.dhis2.org/browse/ANDROAPP-2148)

### DIFF
--- a/app/src/main/java/org/dhis2/Bindings/EventExtensions.kt
+++ b/app/src/main/java/org/dhis2/Bindings/EventExtensions.kt
@@ -1,0 +1,19 @@
+package org.dhis2.Bindings
+
+import org.hisp.dhis.android.core.event.Event
+import org.hisp.dhis.android.core.event.EventStatus
+import java.util.Date
+
+fun Event.primaryDate() : Date {
+
+    return when(status()){
+        EventStatus.ACTIVE -> eventDate()!!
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/dhis2/Bindings/EventExtensions.kt
```diff
@@ -0,0 +1,19 @@
...
+fun Event.primaryDate() : Date {
+
+    return when(status()){
+        EventStatus.ACTIVE -> eventDate()!!
+        EventStatus.COMPLETED -> eventDate()!!
+        EventStatus.SCHEDULE -> dueDate()!!
+        EventStatus.SKIPPED -> dueDate()!!
+        EventStatus.VISITED -> eventDate()!!
+        EventStatus.OVERDUE -> dueDate()!!
+        null -> Date()
```
ferdyrod: What if eventDate() or dueDate() are null?

Balcan: Can't be both null. All Active, completed and visited events have an eventDate. And any schedule, skipped and overdue have a dueDate.
It is a problem in the sdk with wrong @Nullable annotations

Balcan:
> I have tested this issue and it does not working. Do not put the no-sync at the top.

In fact it is working

PR #1053, opened 2019-12-03
fix: Hide Save button during enrollment form loading

JaimeToca:
jira issue : https://jira.dhis2.org/browse/ANDROAPP-2591


Balcan:
Tested! It works :)

PR #1054, opened 2019-12-03
fix(rules): Transform numeric values for RuleDataValues and RuleAttributeValues

Balcan:
- klint reformat.
- Moved tests to correct field.
[jira issue](https://jira.dhis2.org/browse/ANDROAPP-2414)


app/src/main/java/org/dhis2/Bindings/RuleExtensions.kt
```diff
@@ -293,6 +293,8 @@ fun List<TrackedEntityDataValue>.toRuleDataValue(
                     optionRepository.byOptionSetUid().eq(de.optionSetUid()).byCode().eq(value).one()
                         .blockingGet().name()
             }
+        } else if (de.valueType()!!.isNumeric) {
```
JaimeToca: can value type be null if it does not match any value?

Balcan: Nop, all dataElements and trackedEntityAttributes have a valueType. Web does not let you create one without this variable.

PR #1055, opened 2019-12-03
fix(SearchTE): hide map when there is not selected program

mmmateos:
Jira Issue : https://jira.dhis2.org/browse/ANDROAPP-2597


Balcan:
Testing!

PR #1056, opened 2019-12-03
fix(dashboard): Fix crash when removing selected program in landscape dashboard

Balcan:


Balcan:
[jira](https://jira.dhis2.org/browse/ANDROAPP-2598)

mmmateos:
Tested :ok_hand:

PR #1070, opened 2019-12-09
fix: Maps in event programs shows error message when trying to click a second marker

Balcan:
[jira issue](https://jira.dhis2.org/browse/ANDROAPP-2614)
It also has newer version of the SDK (1.0.0) and newer version of the rule engine with a fix for compatibility with android 4.4


mmmateos:
Tested :ok_hand:

JaimeToca:
Hello @Balcan I've just pointed to the master commit since we'll generate an APK from this branch. 

Bitrise gave OKAY to the rule-engine first commit you pushed (the one with JODA time)

PR #1072, opened 2019-12-10
fix(OptionSet): [ANDROAPP-2616] Fix crash when dataSet has an optionSet

mmmateos:


JaimeToca:
@MarinaCape can you check this PR ? Thanks :)

JaimeToca:
tested

PR #1074, opened 2019-12-11
fix(Options): [ANDROAPP-2617] - Fixes order of options in options set

ferdyrod:
**JIRA ISSUEs**: [ANDROAPP-2615](https://jira.dhis2.org/browse/ANDROAPP-2615), [ANDROAPP-2617](https://jira.dhis2.org/browse/ANDROAPP-2617)

### DIFF
--- a/app/src/main/java/org/dhis2/utils/customviews/OptionSetPopUp.java
+++ b/app/src/main/java/org/dhis2/utils/customviews/OptionSetPopUp.java
@@ -23,9 +23,7 @@ import io.reactivex.disposables.CompositeDisposable;
 import io.reactivex.schedulers.Schedulers;
 import timber.log.Timber;
 
-/**
- * QUADRAM. Created by ppajuelo on 20/02/2019.
- */
+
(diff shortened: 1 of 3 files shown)


JaimeToca:
unit test failing :blowfish:

app/src/main/java/org/dhis2/utils/customviews/OptionSetPopUp.java
```diff
@@ -96,6 +95,7 @@ public OptionSetPopUp(Context context, View anchor, SpinnerViewModel model,
 
     public void setOptions(List<Option> options) {
         optionsMap = new HashMap<>();
+        Collections.sort(options, (option1, option2) -> option1.sortOrder().compareTo(option2.sortOrder()));
```
JaimeToca: do you think we could move the sort to another class and unit test it?

ferdyrod: We could, we should discuss it with the team the best way to this re-factorization. We also need to organize the customviews package better. 😕

PR #1076, opened 2019-12-11
fix(Fingerprint): [ANDROAPP-2621] login with fingerprint

mmmateos:
Jira Issue: https://jira.dhis2.org/browse/ANDROAPP-2621


app/src/main/java/org/dhis2/utils/BiometricStorage.java
```diff
@@ -11,23 +14,21 @@
  */
 public class BiometricStorage {
```
JaimeToca: I think you can pass context in the constructor through dagger

JaimeToca:
@mmmateos I think PreferenceProvider should be use instead, right @Balcan ???

PR #1079, opened 2019-12-12
fix: DataSetInstance sync icon not being updated

Balcan:
[jira issue](https://jira.dhis2.org/browse/ANDROAPP-2628)


JaimeToca:
@Balcan what do you think about adding a loading instead?

PR #1081, opened 2019-12-12
fix(DataValuePresenter): Data set display option code and not the name

MarinaCape:
https://jira.dhis2.org/browse/ANDROAPP-2626


ferdyrod:
Tested!

PR #1083, opened 2019-12-13
fix(PeriodDialogInputPeriod): Crash in DataInputPeriod. It is not working

MarinaCape:
https://jira.dhis2.org/browse/ANDROAPP-2613

### DIFF
--- a/app/src/main/java/org/dhis2/utils/customviews/PeriodAdapter.kt
+++ b/app/src/main/java/org/dhis2/utils/customviews/PeriodAdapter.kt
@@ -83,17 +83,7 @@ private class PeriodAdapter(
         var isAllowed = false
         for (inputPeriodModel in inputPeriods) {
             do {
-                if ((currentDate.after(inputPeriodModel.initialPeriodDate()) || currentDate == inputPeriodModel.initialPeriodDate()) && currentDate.before(
-                        inputPeriodModel.endPeriodDate()
-                    )
-                    && (inputPeriodModel.openingDate() == null || inputPeriodModel.openingDate() != null && DateUtils.getInstance().today.after(
-                        inputPeriodModel.openingDate()
-                    )
-                            || DateUtils.getInstance().today == inputPeriodModel.openingDate())
-                    && (inputPeriodModel.closingDate() == null || inputPeriodModel.closingDate() != null && DateUtils.getInstance().today.before(
-                        inputPeriodModel.closingDate()
-                    ))
-                ) {
+                if (isTodayRightDayForInputPeriod(inputPeriodModel)) {
                     isAllowed = true
                     datePeriods.add(currentDate)
                     lastDate = currentDate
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/dhis2/utils/customviews/PeriodAdapter.kt
```diff
@@ -70,5 +79,38 @@ private class PeriodAdapter(
...
+                }
+                else if (currentDate.before(inputPeriods[inputPeriods.size - 1].initialPeriodDate()) || currentDate.before(
+                        inputPeriodModel.initialPeriodDate()
+                    )
+                )
+                    break
+                else {
+                    currentDate = DateUtils.getInstance().getNextPeriod(periodType, currentDate, -1)
+                }
+            } while (!isAllowed)
```
JaimeToca: I think is eassier to follow this code if you change the "do while" to a "for" loop :) 

Please when adding so much logic its better to be in classes that this can be testest, this could have different corner cases that could be identified during unit tests. We need to avoid adding logic to android components :sob:

app/src/main/java/org/dhis2/utils/customviews/PeriodDialogInputPeriod.java
```diff
@@ -65,23 +65,31 @@ else if (getCurrentDate().before(inputPeriod.get(inputPeriod.size() - 1).initial
...
             if (isAllowed) break;
         }
 
         if (!isAllowed && !isEmpty) {
             binding.noPeriods.setText(getString(R.string.there_is_no_available_period));
-        } else {
+        }else{
+            boolean withInputPeriod = false;
+
+            if(isAllowed && !isEmpty){
```
JaimeToca: what is actually "allowed"?

app/src/main/java/org/dhis2/utils/customviews/PeriodAdapter.kt
```diff
@@ -70,5 +79,38 @@ private class PeriodAdapter(
...
+                    ))
+                ) {
+                    isAllowed = true
+                    datePeriods.add(currentDate)
+                    lastDate = currentDate
+                }
+                else if (currentDate.before(inputPeriods[inputPeriods.size - 1].initialPeriodDate()) || currentDate.before(
+                        inputPeriodModel.initialPeriodDate()
+                    )
+                )
```
JaimeToca: not sure if this pass the klint I believe it will tell you need to add "{"

app/src/main/java/org/dhis2/utils/customviews/PeriodAdapter.kt
```diff
@@ -70,5 +79,38 @@ private class PeriodAdapter(
...
+                if ((currentDate.after(inputPeriodModel.initialPeriodDate()) || currentDate == inputPeriodModel.initialPeriodDate()) && currentDate.before(
+                        inputPeriodModel.endPeriodDate()
+                    )
+                    && (inputPeriodModel.openingDate() == null || inputPeriodModel.openingDate() != null && DateUtils.getInstance().today.after(
+                        inputPeriodModel.openingDate()
+                    )
+                            || DateUtils.getInstance().today == inputPeriodModel.openingDate())
+                    && (inputPeriodModel.closingDate() == null || inputPeriodModel.closingDate() != null && DateUtils.getInstance().today.before(
+                        inputPeriodModel.closingDate()
+                    ))
```
JaimeToca: I think readability improves a bit if you do this :  

fun isTodayBetweenInitialAndEndPeriod() : Boolean{
return ...
}
fun isTodayBetweenOpenAndCloseDate() :  Boolean{
return ...
}

And then

if (isTodayBetweenInitialAndEndPeriod() &&  isTodayBetweenOpenAndCloseDate()) {
                    isAllowed = true
                    datePeriods.add(currentDate)
                    lastDate = currentDate
}

MarinaCape: You're rigth!

PR #1084, opened 2019-12-13
fix(EnrollmentForm): [ANDROAPP-2631] - handles TEI coordinate field in servers 2.29 and 2.30

ferdyrod:
**JIRA ISSUES**: [ANDROAPP-2631](https://jira.dhis2.org/browse/ANDROAPP-2631)

### DIFF
--- a/app/src/main/java/org/dhis2/usescases/enrollment/EnrollmentActivity.kt
+++ b/app/src/main/java/org/dhis2/usescases/enrollment/EnrollmentActivity.kt
@@ -499,8 +499,9 @@ class EnrollmentActivity : ActivityGlobalAbstract(), EnrollmentView {
     override fun displayTeiCoordinates(
         teiCoordinatesData: Pair<TrackedEntityType, TrackedEntityInstance>?
     ) {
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/dhis2/usescases/enrollment/EnrollmentActivity.kt
```diff
@@ -499,8 +499,9 @@ class EnrollmentActivity : ActivityGlobalAbstract(), EnrollmentView {
     override fun displayTeiCoordinates(
         teiCoordinatesData: Pair<TrackedEntityType, TrackedEntityInstance>?
     ) {
+        val featureType = teiCoordinatesData!!.first.featureType() ?: FeatureType.NONE
```
MarinaCape: Could teiCoordinatesData be null sometime?

ferdyrod: No it could not. `teiCoordinatesData` is a pair of TrackedEntityType and TrackedEntityInstance. Every TrackedEntityInstance will have a TrackedEntityType.

PR #1092, opened 2019-12-18
test: adds test on programEventDetailPresenter

mmmateos:
init method should be reviewed to add all test


JaimeToca:
@mmmateos this PR can point to develop :)

PR #1094, opened 2019-12-18
fix: coordinates callback

juanfranem:


JaimeToca:
to me everything fine. @MarinaCape @juanfranem  @ferdyrod can you make sure sdk submodule and rule engine are okay? Sdk should point to develop, rule engine to master

app/src/main/java/org/dhis2/usescases/enrollment/EnrollmentActivity.kt
```diff
@@ -92,6 +97,22 @@ class EnrollmentActivity : ActivityGlobalAbstract(), EnrollmentView {
 
     /*region LIFECYCLE*/
 
+
+    override fun onRequestPermissionsResult(requestCode: Int, permissions: Array<out String>, grantResults: IntArray) {
+        when (requestCode) {
+            EventInitialPresenter.ACCESS_COARSE_LOCATION_PERMISSION_REQUEST -> {
+                // If request is cancelled, the result arrays are empty.
+                if (grantResults.isNotEmpty() && grantResults[0] == PackageManager.PERMISSION_GRANTED) {
+                    coordinatesView?.let {
```
Balcan: The enrollment activity has 2 coordinates views. Why is there only one being used here?

PR #1098, opened 2019-12-19
fix: changing apk signing to bitrise

JaimeToca:


JaimeToca:
@Balcan I'm removing this PR since it's done in 2.0.0. We'll merge 2.0.0 into master and develop then this will be include....

PR #1119, opened 2020-01-09
feat: adds informative messages when sync with SMS

mmmateos:
Jira issue: https://jira.dhis2.org/browse/ANDROAPP-2638


ferdyrod:
This was tested on device!!!

PR #1121, opened 2020-01-13
fix: I set, ask pin everytime the app restarts

Balcan:


app/src/main/java/org/dhis2/usescases/login/LoginActivity.kt
```diff
@@ -91,11 +91,11 @@ class LoginActivity : ActivityGlobalAbstract(), LoginContracts.View {
...
-        binding.pinLayout.forgotCode.visibility = View.VISIBLE
-        binding.pinLayout.forgotCode.setOnClickListener {
-            analyticsHelper.setEvent(FORGOT_CODE, CLICK, FORGOT_CODE)
-            binding.pinLayout.root.visibility = View.GONE
-        }
+        /* binding.pinLayout.forgotCode.visibility = View.VISIBLE
+         binding.pinLayout.forgotCode.setOnClickListener {
+             analyticsHelper.setEvent(FORGOT_CODE, CLICK, FORGOT_CODE)
+             binding.pinLayout.root.visibility = View.GONE
+         }*/
```
ferdyrod: If this is not needed we should eliminate this piece of code ☺️

app/src/main/java/org/dhis2/usescases/login/LoginActivity.kt
```diff
@@ -289,22 +289,32 @@ class LoginActivity : ActivityGlobalAbstract(), LoginContracts.View {
...
+
+             override fun onEmpty() {
+             }
+
+             override fun onPinChange(pinLength: Int, intermediatePin: String) {
+             }
+         })
+         binding.pinLayout.title.text = getString(R.string.unblock_session)
+         binding.pinLayout.root.visibility = View.VISIBLE
+         isPinScreenVisible = true*/
```
ferdyrod: Commented code

app/src/main/java/org/dhis2/usescases/main/MainActivity.kt
```diff
@@ -85,7 +87,7 @@ class MainActivity : ActivityGlobalAbstract(), MainView, ExporterListener {
             override fun onEmpty() {}
 
             override fun onPinChange(pinLength: Int, intermediatePin: String) {}
-        })
+        })*/
```
ferdyrod: Commented Code

app/src/main/java/org/dhis2/utils/session/PinDialog.kt
```diff
@@ -0,0 +1,110 @@
...
+        super.onCreate(savedInstanceState)
+        setStyle(STYLE_NO_TITLE, android.R.style.Theme_DeviceDefault_Light_NoActionBar)
+        app().createSessionComponent(PinModule(this)).inject(this)
+    }
+
+    override fun onCreateDialog(savedInstanceState: Bundle?): Dialog {
+        val dialog = super.onCreateDialog(savedInstanceState)
+        dialog.window!!.requestFeature(Window.FEATURE_NO_TITLE)
+        dialog.window!!.setBackgroundDrawableResource(android.R.color.transparent)
+        dialog.window!!.setWindowAnimations(R.style.pin_dialog_animation)
```
ferdyrod: we can use _apply_ or _let_ high-order function in here

ferdyrod:
When testing this PR, I found out that if the app is sent to background it does not ask por the pin when putting the app in the foreground.

JaimeToca:
Please @Balcan check out rule engine

PR #1122, opened 2020-01-14
docs: Create new folder structure to support more docs [skip ci]

jaboto:
I have changed the whole doc structure to support the addition of new documents related to the Application but not the Application per se. 

Another PR has been requested on the dhis2-docs repository so the generation of the documentation takes these changes into account.


jaboto:
A file was missing... will re PR

PR #1134, opened 2020-01-15
fix: [ANDROAPP-2650] Check if value is correct before getting/saving it

Balcan:
[jira issue](https://jira.dhis2.org/browse/ANDROAPP-2650)


app/src/main/java/org/dhis2/Bindings/ValueExtensions.kt
```diff
@@ -3,27 +3,158 @@ package org.dhis2.Bindings
...
+        .byCode().eq(code).one().blockingGet()?.displayName()
+}
+
+fun checkValueTypeValue(d2: D2, valueType: ValueType?, value: String): String {
+    return when (valueType) {
+        ValueType.ORGANISATION_UNIT ->
+            d2.organisationUnitModule().organisationUnits()
+                .uid(value)
+                .blockingGet()
+                .displayName()!!
```
JaimeToca: Can display name be null ? If so it will crash (nobody catch the exception)

app/src/main/java/org/dhis2/Bindings/ValueExtensions.kt
```diff
@@ -3,27 +3,158 @@ package org.dhis2.Bindings
...
+        ValueType.IMAGE, ValueType.FILE_RESOURCE ->
+            if (d2.fileResourceModule().fileResources().uid(value).blockingExists()) {
+                d2.fileResourceModule().fileResources().uid(value).blockingGet().path()!!
```
JaimeToca: same for path

app/src/main/java/org/dhis2/Bindings/ValueExtensions.kt
Balcan: Nop, displayName is never null, this may be one of the wrong nullable tags in the sdk models

app/src/main/java/org/dhis2/Bindings/ValueExtensions.kt
Balcan: Same answer :)

ferdyrod:
This was tested on devices and is working fine!

PR #1135, opened 2020-01-16
docs: Implementation guideline document [skip ci]

jaboto:
Initial release of the document migrated from Google doc.

To be completed:
* Links are missing
* Review image size
* Review whole layout


jaboto:
This document will not be show available on the DHIS 2 documentation portal until it is fully ready. Then, a PR will be make to dhis2-docs repo to include the INDEX file.

PR #1137, opened 2020-01-17
feat: [ANDROAPP-1670] adds viewModel for barcode/qr

mmmateos:
Jira Issue: https://jira.dhis2.org/browse/ANDROAPP-1670


app/src/main/java/org/dhis2/usescases/eventsWithoutRegistration/eventCapture/EventCaptureActivity.java
```diff
@@ -164,6 +164,10 @@ protected void onActivityResult(int requestCode, int resultCode, Intent data) {
                     presenter.nextCalculation(true);
                 }
                 break;
+            case Constants.RQ_QR_SCANNER:
+                if(resultCode == RESULT_OK) {
+                    scanTextView.updateScanResult(data.getStringExtra(Constants.EXTRA_DATA));
+                }
```
JaimeToca: I would add break; just in case we add another thing here in the future we avoid strange behaviours

app/src/main/java/org/dhis2/utils/customviews/ScanTextView.kt
```diff
@@ -0,0 +1,146 @@
...
+        init(context)
+    }
+
+    constructor(context: Context, attrs: AttributeSet, defStyle: Int) : super(
+        context,
+        attrs,
+        defStyle
+    ) {
+        init(context)
+    }
```
ferdyrod: There is a easier way to achieve all of this in Kotlin with very few lines of code.

``` kotlin
class ScanTestView @JvmOverloads constructor(
    context: Context,
    attrs: AttributeSet? = null,
    defStyle: Int = 0
) : FieldLayout(context, attrs, defStyle) {

    private lateinit var iconView: ImageView
    ...
    private lateinit var onScanResult: (String?) -> Unit
    
    init {
        init(context)
    }
}
````

the `@JvmOverloads` will create all the different constructors depending on the defaults parameter of the main constructor. 

Then we use the `init` block to call the FieldLayout.init(context) method when initializing this class.

File: app/src/main/java/org/dhis2/usescases/qrScanner/QRActivity.java [hunk 3]
```diff
@@ -43,6 +43,7 @@ protected void onCreate(@Nullable Bundle savedInstanceState) {
         mScannerView.setAutoFocus(true);
         ArrayList<BarcodeFormat> formats = new ArrayList<>();
         formats.add(BarcodeFormat.QR_CODE);
+        formats.add(BarcodeFormat.CODE_128);
```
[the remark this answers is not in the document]
Balcan: We need all of them

Balcan:
Tested

PR #1139, opened 2020-01-20
fix: [ANDROAPP-2650] Fixed remaining value checks that blocked data entry

Balcan:
[jira issue](https://jira.dhis2.org/browse/ANDROAPP-2650)


JaimeToca:
conflicts

PR #1140, opened 2020-01-20
fix: [ANDROAPP-2659] don't search if all fields are empty

mmmateos:
Jira Issue: https://jira.dhis2.org/browse/ANDROAPP-2659


ferdyrod:
Tested!

PR #1145, opened 2020-01-23
fix: [ANDROAPP-2664] Filter out active events with dates in the future

Balcan:
[jira issue](https://jira.dhis2.org/browse/ANDROAPP-2664)


app/src/test/java/android/text/TextUtils.java
```diff
@@ -0,0 +1,7 @@
+package android.text;
+
+public class TextUtils {
+    public static boolean isEmpty(CharSequence str) {
+        return str == null || str.length() == 0;
+    }
+}
```
JaimeToca: I think we have DhisTextUtils :D

PR #1155, opened 2020-01-29
docs: transform html tables to GFM [skip ci]

jaboto:
Transformed HTML tables to GFM where possible


jaboto:
@Philip-Larsen-Donnelly I have updated the tables where possible. However there is a specific table that cannot be converted as we use colspan (check file: https://github.com/dhis2/dhis2-android-capture-app/blob/master/docs/src/commonmark/en/content/implementation-guide/mobile-device-specs.md ) which seems to be render well in Github and in the HTML and PDF generated with the build.sh.

Please let me know otherwise

PR #1156, opened 2020-01-29
fix: [ANDROAPP-2680] view and create notes when enrollment is complet…

mmmateos:
…e/deactivate

Jira Issue: https://jira.dhis2.org/browse/ANDROAPP-2680


app/src/main/java/org/dhis2/usescases/notes/NotesFragment.kt
```diff
@@ -116,15 +116,21 @@ class NotesFragment : FragmentGlobalAbstract(), NotesView {
...
-        noteAdapter.setItems(noteModules)
+    override fun swapNotes(notes: List<Note>) {
+        if (notes.isEmpty()) {
+            binding.noNotesLayout.visibility = View.VISIBLE
+            binding.notesRecycler.visibility = View.GONE
+        } else {
+            binding.noNotesLayout.visibility = View.GONE
+            binding.notesRecycler.visibility = View.VISIBLE
+            noteAdapter.setItems(notes)
+        }
```
JaimeToca: I think this logic could go in the presenter and call methods view.showNotes or not . It's small but doing this we put less responsability in view and can be unit test.

PR #1160, opened 2020-01-30
fix: Adapt SDK changes for database after log in

JaimeToca:


JaimeToca:
I saw that the method biometric button does not have unit test. I'll try to add it !

PR #1169, opened 2020-02-04
fix: [ANDROAPP-1670] Updated search activity to be able to use QR/Barcode scanner

Balcan:
[jira](https://jira.dhis2.org/browse/ANDROAPP-1670)


JaimeToca:
code review and test. since @mmmateos is on holiday and ferdy was busy. I think it looks good to me

PR #1176, opened 2020-02-06
fix: [ANDROAPP-2727] Fixed loop in event form after completing the event 

Balcan:
[jira](https://jira.dhis2.org/browse/ANDROAPP-2727)


app/src/main/java/org/dhis2/usescases/eventsWithoutRegistration/eventCapture/EventCaptureRepositoryImpl.java
```diff
@@ -644,7 +644,8 @@ public String getSectionFor(String field) {
     @Override
     public Flowable<Boolean> eventIntegrityCheck() {
         return d2.eventModule().events().uid(eventUid).get()
-                .map(event -> event.status() == EventStatus.ACTIVE && event.eventDate() != null && !event.eventDate().after(new Date()))
+                .map(event -> (event.status() == EventStatus.ACTIVE ||
+                        event.status() == EventStatus.COMPLETED) && event.eventDate() != null && !event.eventDate().after(new Date()))
                 .toFlowable();
```
JaimeToca: I missed a unit test for this :( . For this small bugs having something to make sure it never happened again

JaimeToca: ok, I seed you coudln't due to contex usage. 

The same as preferences @Balcan , we should have a provider to encapsulate context.getString(R.string.enter_text) this :) . If you can add it in our things to :+1:

JaimeToca:
@Balcan I think you have conflicts. Can you check my comment too :)

app/src/main/java/org/dhis2/usescases/eventsWithoutRegistration/eventCapture/EventCaptureRepositoryImpl.java
Balcan: I'll add it. Moreover, this class requieres mocking a lot of calls to d2. It will take quite some time to get the test class ready

PR #1178, opened 2020-02-06
fix: [Androapp 2725] Fix percentage calculation by removing display elements

Balcan:
[jira](https://jira.dhis2.org/browse/ANDROAPP-2725)

### DIFF
--- a/app/src/main/java/org/dhis2/usescases/eventsWithoutRegistration/eventCapture/EventCapturePresenterImpl.java
+++ b/app/src/main/java/org/dhis2/usescases/eventsWithoutRegistration/eventCapture/EventCapturePresenterImpl.java
@@ -251,6 +251,16 @@ public class EventCapturePresenterImpl implements EventCaptureContract.Presenter
         compositeDisposable.add(
                 eventCaptureRepository.eventSections()
                         .flatMap(sectionList -> fieldFlowable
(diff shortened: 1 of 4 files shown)


app/src/main/java/org/dhis2/usescases/eventsWithoutRegistration/eventCapture/EventCapturePresenterImpl.java
```diff
@@ -251,6 +251,16 @@ public void init() {
...
+                                .map(fields->{
+                                    Iterator<FieldViewModel> iterator = fields.iterator();
+                                    while (iterator.hasNext()){
+                                        FieldViewModel field = iterator.next();
+                                        if(field instanceof DisplayViewModel){
+                                            iterator.remove();
+                                        }
+                                    }
+                                    return fields;
+                                })
```
JaimeToca: why not avoid adding these fields instead of removing them? it's not very clear why in this part of the code we are removing them

Balcan: Cause all these fields are needed for the forms. The source is the same for both showing the form fields and calculating the percentage. But for the later, you have to remove the display fields created by program rules.

PR #1185, opened 2020-02-13
fix: [ANDROAPP-2740] crash in relationship

mmmateos:
Jira Issue: https://jira.dhis2.org/browse/ANDROAPP-2740


ferdyrod:
Tested!

PR #1186, opened 2020-02-14
fix: [ANDROAPP-2739] adjusting form message on notes tab

mmmateos:
Jira Issue: https://jira.dhis2.org/browse/ANDROAPP-2739


ferdyrod:
We are closing this PR because there is a new design to the forms which includes this fix.

Fix in PR #1187

PR #1188, opened 2020-02-18
fix: [ANDROAPP-2737] Value on a text field is saved when dismissing the Keyboard

ferdyrod:
**JIRA ISSUE**: [ANDROAPP-2737](https://jira.dhis2.org/browse/ANDROAPP-2737)


app/src/main/java/org/dhis2/utils/customviews/TextInputAutoCompleteTextView.java
```diff
@@ -44,4 +41,13 @@ public InputConnection onCreateInputConnection(EditorInfo outAttrs) {
         }
         return ic;
     }
+
+
+    @Override
+    public boolean onKeyPreIme(int keyCode, KeyEvent event) {
+        if(keyCode == KeyEvent.KEYCODE_BACK){
```
JaimeToca: can you move this into a method, so at least we can remember in 6 mothns why is this here.

fun clearFocusToSaveInfoWhenGoingBack(){

}

or something like this... thoughts?

ferdyrod: @JaimeToca For something so simple, I believe a Javadoc style comment is sufficient, right? Don't see the value of moving this to another method.

JaimeToca: Always better if you try to explain yourself in the code. I know this is not the case. But when you have code that you need to comment, it's a smell and it means that you need to rename variables/methods names or structure better. To the developer is always easier to read the code than comments. Most of the time comments are not updated with the code changes.

mmmateos:
Not working when back arrow is clicked

PR #1195, opened 2020-02-21
fix: [ANDROAPP-2758]: not showing mandatory fields alert

mmmateos:
Jira Issue: https://jira.dhis2.org/browse/ANDROAPP-2758


Balcan:
DataEntryAdapter has changed and now the viewModels are not being saved. So this fix must be discarded

PR #1199, opened 2020-02-25
fix: [ANDROAPP-2743] Grey out data sets when "Assigned to me" filter is active

Balcan:
[issue](https://jira.dhis2.org/browse/ANDROAPP-2743)


app/src/test/java/org/dhis2/usescases/main/program/HomeRepositoryImplTest.kt
```diff
@@ -0,0 +1,283 @@
+package org.dhis2.usescases.main.program
+
+import com.nhaarman.mockitokotlin2.doReturn
+import com.nhaarman.mockitokotlin2.mock
+import com.nhaarman.mockitokotlin2.times
+import com.nhaarman.mockitokotlin2.verify
```
JaimeToca: GREAT

PR #1200, opened 2020-02-25
fix: [ANDROAPP-2738] fixes deletion of a Tei or an Enrollement

ferdyrod:
**JIRA ISSUE**: [ANDROAPP-2738](https://jira.dhis2.org/browse/ANDROAPP-2738)

### DIFF
--- a/app/src/main/java/org/dhis2/usescases/searchTrackEntity/SearchRepositoryImpl.java
+++ b/app/src/main/java/org/dhis2/usescases/searchTrackEntity/SearchRepositoryImpl.java
@@ -13,6 +13,7 @@ import androidx.paging.PagedList;
 import com.squareup.sqlbrite2.BriteDatabase;
 
 import org.dhis2.Bindings.ExtensionsKt;
+import org.dhis2.Bindings.TrackedEntityInstanceExtensionsKt;
 import org.dhis2.Bindings.ValueExtensionsKt;
 import org.dhis2.data.tuples.Pair;
 import org.dhis2.data.tuples.Trio;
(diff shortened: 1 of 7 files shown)


app/src/main/java/org/dhis2/usescases/searchTrackEntity/SearchRepositoryImpl.java
```diff
@@ -589,6 +598,22 @@ else if (d2.trackedEntityModule().trackedEntityAttributes().byUid().eq(tetAttrib
...
+                if (program != null) {
+                    Enrollment enrollment = d2.enrollmentModule().enrollments().byTrackedEntityInstance().eq(tei.uid()).byProgram().eq(program.uid()).one().blockingGet();
+                    if (enrollment.program().equals(program.uid()) && enrollment.deleted()) {
+                        iterator.remove();
+                    }
+                }
+            }
+        }
+        return teis;
+    }
```
JaimeToca: to me this should go outside repository, it's not its responsability

JaimeToca: thoughts? @ferdyrod

ferdyrod: Yes, it should. this whole repository needs a lot of rework.

JaimeToca: :+1:  great ! do you think this small piece can be extracted (only filterDeletedEnrollment) or has other considerations??

Balcan: I'm going to extract it to an extension function

PR #1204, opened 2020-02-27
fix: [ANDROAPP-2772] Fix loop with program rules action assign and error

Balcan:
[issue](https://jira.dhis2.org/browse/ANDROAPP-2772)
- Enrollment count fix

### DIFF
--- a/app/src/main/java/org/dhis2/usescases/eventsWithoutRegistration/eventCapture/EventCapturePresenterImpl.java
+++ b/app/src/main/java/org/dhis2/usescases/eventsWithoutRegistration/eventCapture/EventCapturePresenterImpl.java
@@ -492,21 +492,7 @@ public class EventCapturePresenterImpl implements EventCaptureContract.Presenter
             Timber.e(calcResult.error());
             return viewModels;
         }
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/dhis2/usescases/eventsWithoutRegistration/eventCapture/EventCapturePresenterImpl.java
```diff
@@ -488,6 +492,20 @@ public void nextCalculation(boolean doNextCalculation) {
...
+            if(action instanceof RuleActionAssign){
+                if(((RuleActionAssign) action).field().startsWith("JGQ")||((RuleActionAssign) action).field().startsWith("gDe")){
+                    testRuleEffect.add(effect);
+                }
+            }else if(action instanceof RuleActionHideField){
+                if(((RuleActionHideField) action).field().startsWith("JGQ")||((RuleActionHideField) action).field().startsWith("gDe")){
+                    testRuleEffect.add(effect);
+                }
+            }
+        }
```
JaimeToca: It's a problem from the rule engine??? I am worry about the fact that they can have the same problem with another field different than "JGQ" or "gDe" ??

PR #1209, opened 2020-03-03
fix: [ANDROAPP-2770] Disable settings parameter when there is no internet connection

Balcan:
[jira issue](https://jira.dhis2.org/browse/ANDROAPP-2770)

### DIFF
--- a/app/src/main/java/org/dhis2/usescases/settings/SyncManagerFragment.java
+++ b/app/src/main/java/org/dhis2/usescases/settings/SyncManagerFragment.java
@@ -3,7 +3,10 @@ package org.dhis2.usescases.settings;
 import android.Manifest;
 import android.app.NotificationChannel;
 import android.app.NotificationManager;
+import android.content.BroadcastReceiver;
 import android.content.Context;
+import android.content.Intent;
+import android.content.IntentFilter;
 import android.content.SharedPreferences;
 import android.content.pm.PackageManager;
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/dhis2/usescases/settings/SyncManagerFragment.java
```diff
@@ -115,6 +124,7 @@ public View onCreateView(@NonNull LayoutInflater inflater, ViewGroup container,
     @Override
     public void onResume() {
         super.onResume();
+        context.registerReceiver(networkReceiver, new IntentFilter("android.net.conn.CONNECTIVITY_CHANGE"));
```
JaimeToca: I think this was deprecated in 28 
https://developer.android.com/reference/android/net/ConnectivityManager#CONNECTIVITY_ACTION

JaimeToca: the value of the constant action is actually Constant Value: "android.net.conn.CONNECTIVITY_CHANGE"

Balcan: In fact it is the registration of the receiver in the manifest what is deprecated. It is valid to use if it is registered programmatically using a valid context.

ferdyrod:
Tested!

PR #1219, opened 2020-03-05
fix: [ANDROAPP-2757] add sticky header on events

mmmateos:
[https://jira.dhis2.org/browse/ANDROAPP-2757](Jira Issue)


app/src/main/java/org/dhis2/usescases/eventsWithoutRegistration/eventCapture/EventCaptureFragment/EventCaptureFormPresenter.kt
```diff
@@ -60,12 +62,27 @@ class EventCaptureFormPresenter(
...
                 .observeOn(schedulerProvider.ui())
                 .subscribe(
-                    { view.showFields(it, lastFocusItem) },
+                    { fields ->
+                        view.showFields(fields, lastFocusItem)
+                        selectedSection ?: fields
+                            .mapNotNull { it.programStageSection() }
+                            .firstOrNull()
+                            ?.let{ selectedSection = it }
+                    },
```
JaimeToca: Event capture form presenter has already unit test for this. I think it would be very simple and worthy to try to add some checks to this new behaviour

PR #1222, opened 2020-03-06
fix: [ANDROAPP-2788 | ANDROAPP-2742] Attribute value for rule variable is always empty

Balcan:
[issue](https://jira.dhis2.org/browse/ANDROAPP-2788)


JaimeToca:
@Balcan Unit test failing

PR #1226, opened 2020-03-09
fix: [ANDROAPP-2729] Location using map in search fields

ferdyrod:
**JIRA ISSUE** [ANDROAPP-2729] (https://jira.dhis2.org/browse/ANDROAPP-2729)


JaimeToca:
@ferdyrod  pls make sure there is info for testing regarding, other fields, qr etc... since we had issue with that in the past. Thanks!

PR #1230, opened 2020-03-10
fix: [ANDROAPP-1670] scan only options of optionSet

mmmateos:
https://jira.dhis2.org/browse/ANDROAPP-1670


app/src/main/java/org/dhis2/usescases/qrScanner/ScanActivity.kt
```diff
@@ -0,0 +1,103 @@
...
+        mScannerView.startCamera()
+    }
+
+    override fun onRequestPermissionsResult(
+        requestCode: Int,
+        permissions: Array<String>,
+        grantResults: IntArray
+    ) {
+        when (requestCode) {
+            101 -> {
```
JaimeToca: can you use a constant?

app/src/main/java/org/dhis2/usescases/qrScanner/ScanActivity.kt
```diff
@@ -0,0 +1,103 @@
...
+
+    override fun onCreate(savedInstanceState: Bundle?) {
+        super.onCreate(savedInstanceState)
+        optionSetUid = intent.getStringExtra(Constants.OPTION_SET)
+        (applicationContext as App)
+            .userComponent()
+            ?.plus(ScanModule(optionSetUid))
+            ?.inject(this)
+
+        binding = DataBindingUtil.setContentView(this, R.layout.activity_qr)
```
ferdyrod: Maybe it is a good idea to also change the name of the layout to correspond to the activity's name? What do you think?

app/src/main/java/org/dhis2/usescases/qrScanner/ScanActivity.kt
mmmateos: Sure!

app/src/main/java/org/dhis2/usescases/qrScanner/ScanActivity.kt
mmmateos: You are right

PR #1245, opened 2020-03-13
fix: [ANDROAPP-2813] Show full label in every field

Balcan:
[issue](https://jira.dhis2.org/browse/ANDROAPP-2813)


app/src/main/java/org/dhis2/Bindings/ValueTypeExtensions.kt
```diff
@@ -0,0 +1,41 @@
...
+        ValueType.DATETIME,
+        ValueType.TIME -> context.getString(R.string.enter_value)
+    }
+}
+
+fun Context.valueTypeHintMap(): Map<ValueType, String> {
+    return ValueType.values().map {
+        it to it.toHint(this)
+    }.toMap()
+}
```
ferdyrod: This looks like a context extension rather than a valueType extension.

Balcan: But it is heavily related to the ValueType. That's why I decided to leave it there.

app/src/main/java/org/dhis2/Bindings/ValueTypeExtensions.kt
```diff
@@ -0,0 +1,41 @@
...
+        ValueType.DATE,
+        ValueType.DATETIME,
+        ValueType.TIME -> context.getString(R.string.enter_value)
+    }
+}
+
+fun Context.valueTypeHintMap(): Map<ValueType, String> {
+    return ValueType.values().map {
+        it to it.toHint(this)
+    }.toMap()
```
JaimeToca: now having a map, could we get a crash when doing .get , if it does not match it?

Balcan: Nop, cause it is using its own values

JaimeToca:
could you add to the ticket, to test all the parts that might have side effects :pray:  thanks !

PR #1257, opened 2020-03-16
fix: [ANDROAPP-2818] Updates location of map when granting location permission.

ferdyrod:
**JIRA ISSUE**: [ANDROAPP-2818](https://jira.dhis2.org/browse/ANDROAPP-2818)

### DIFF
--- a/app/src/main/java/org/dhis2/usescases/map/MapSelectorActivity.kt
+++ b/app/src/main/java/org/dhis2/usescases/map/MapSelectorActivity.kt
@@ -45,10 +45,6 @@ import org.hisp.dhis.android.core.common.FeatureType
 import org.hisp.dhis.android.core.common.Geometry
 import timber.log.Timber
 
-/**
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/dhis2/usescases/map/MapSelectorActivity.kt
```diff
@@ -386,32 +382,17 @@ class MapSelectorActivity :
...
         ) {
             // Should we show an explanation?
-            /*if (ActivityCompat.shouldShowRequestPermissionRationale(this, Manifest.permission.ACCESS_COARSE_LOCATION)) {
-                // TODO CRIS
-                // Show an expanation to the user *asynchronously* -- don't block
-                // this thread waiting for the user's response! After the user
-                // sees the explanation, try again to request the permission.
-
-            }*/
+            // TODO CRIS
```
Balcan: I think you can remove all these comments

PR #1258, opened 2020-03-16
fix: [ANDROAPP-2819] check enrollment followUp is not null

mmmateos:
[Jira Issue](https://jira.dhis2.org/browse/ANDROAPP-2819)


app/src/main/java/org/dhis2/usescases/teiDashboard/dashboardfragments/tei_data/TEIDataFragment.java
```diff
@@ -214,7 +214,7 @@ public void onPause() {
     public void setEnrollmentData(Program program, Enrollment enrollment) {
         binding.setProgram(program);
         binding.setEnrollment(enrollment);
-        if (enrollment != null) {
+        if (enrollment != null && enrollment.followUp() != null) {
```
Balcan: If the enrollment is not null, a null follow up should be treated as false just in case.

PR #1266, opened 2020-03-17
style: [ANDROAPP-2821] Fix enrollment toolbar title

Balcan:
[issue](https://jira.dhis2.org/browse/ANDROAPP-2821)

### DIFF
--- a/app/src/main/java/org/dhis2/usescases/enrollment/EnrollmentActivity.kt
+++ b/app/src/main/java/org/dhis2/usescases/enrollment/EnrollmentActivity.kt
@@ -340,8 +340,17 @@ class EnrollmentActivity : ActivityGlobalAbstract(), EnrollmentView {
             binding.teiDataHeader.mainAttributes.text =
                 String.format("%s %s", firstAttr, secondAttr)
             binding.teiDataHeader.secundaryAttribute.text = thirdAttr
+
+            if(firstAttr.isEmpty() && secondAttr.isEmpty()){
+                binding.teiDataHeader.mainAttributes.visibility = View.GONE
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/dhis2/usescases/enrollment/EnrollmentActivity.kt
```diff
@@ -340,8 +340,17 @@ class EnrollmentActivity : ActivityGlobalAbstract(), EnrollmentView {
             binding.teiDataHeader.mainAttributes.text =
                 String.format("%s %s", firstAttr, secondAttr)
             binding.teiDataHeader.secundaryAttribute.text = thirdAttr
+
+            if(firstAttr.isEmpty() && secondAttr.isEmpty()){
+                binding.teiDataHeader.mainAttributes.visibility = View.GONE
+            }
+            if(thirdAttr.isEmpty()){
+                binding.teiDataHeader.secundaryAttribute.visibility = View.GONE
+            }
```
JaimeToca: this is small but I usually would have put it in the presenter method that handles the view logic.

Balcan: That will make it a bigger change. So maybe we can do it next time

JaimeToca: I had approve it the PR. Won't do it next time...

The logic of the view should always go to the presenter as a norm, that is the presenter responsibility. You could have 2 methods in the activity and call them from the presenter. In addition, you could have your happy unit tests.

PR #1280, opened 2020-03-19
fix: [ANDROAPP-2831] Sets elevation when filters are shown in the TEI Dashboard

ferdyrod:
**JIRA ISSUE**: [ANDROAPP-2831](https://jira.dhis2.org/browse/ANDROAPP-2831)

### DIFF
--- a/app/src/main/java/org/dhis2/usescases/teiDashboard/TeiDashboardMobileActivity.java
+++ b/app/src/main/java/org/dhis2/usescases/teiDashboard/TeiDashboardMobileActivity.java
@@ -17,6 +17,7 @@ import android.widget.PopupMenu;
 
 import androidx.annotation.Nullable;
 import androidx.core.content.ContextCompat;
+import androidx.core.view.ViewCompat;
 import androidx.databinding.DataBindingUtil;
 import androidx.lifecycle.LiveData;
 import androidx.lifecycle.MutableLiveData;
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/dhis2/usescases/teiDashboard/TeiDashboardMobileActivity.java
```diff
@@ -647,12 +651,14 @@ public void updateTotalFilters(Integer totalFilters) {
...
         binding.tabLayout.setVisibility(View.GONE);
         binding.teiPager.setUserInputEnabled(false);
+        ViewCompat.setElevation(binding.toolbar, 0);
     }
 
     @Override
     public void showTabsAndEnableSwipe() {
         binding.tabLayout.setVisibility(View.VISIBLE);
         binding.teiPager.setUserInputEnabled(true);
+        ViewCompat.setElevation(binding.toolbar, elevation);
```
Balcan: Why elevation is always 0? Shouldn't it be set to the default value?

ferdyrod: Yes,  in OnCreate() I get the default value of the elevation of the toolbar. Here I set that value to the default. The elevation variable is never set to 0

PR #1281, opened 2020-03-19
fix: [ANDROAPP-2832] section error indicators

mmmateos:
[Jira Issue](https://jira.dhis2.org/browse/ANDROAPP-2832)


ferdyrod:
Tested!

PR #1282, opened 2020-03-20
fix: [ANDROAPP-2838] Do not delete values when hide section is activated

Balcan:
[issue](https://jira.dhis2.org/browse/ANDROAPP-2838)


ferdyrod:
Tested!

PR #1283, opened 2020-03-20
fix: [ANDROAPP-2835] Display full image when clicked

Balcan:
[issue](https://jira.dhis2.org/browse/ANDROAPP-2839)


ferdyrod:
Tested

PR #1284, opened 2020-03-23
fix: [ANDROAPP-2839] treat displayViewModels as unsupported fields in…

mmmateos:
… completion percentage

[Jira Issue](https://jira.dhis2.org/browse/ANDROAPP-2839)


JaimeToca:
is it difficult to add unit test @mmmateos ?

PR #1286, opened 2020-03-24
fix: [ANDROAPP-2815] Avoid multiple requests to tei query api

Balcan:
[issue](https://jira.dhis2.org/browse/ANDROAPP-2815)


JaimeToca:
Perhas this should point to develop @Balcan ??

PR #1291, opened 2020-03-25
fix: [ANDROAPP-2846] TEI details title attributes

mmmateos:
[Jira Issue](https://jira.dhis2.org/browse/ANDROAPP-2846)

### DIFF
--- a/app/src/main/java/org/dhis2/usescases/enrollment/EnrollmentActivity.kt
+++ b/app/src/main/java/org/dhis2/usescases/enrollment/EnrollmentActivity.kt
@@ -316,29 +316,22 @@ class EnrollmentActivity : ActivityGlobalAbstract(), EnrollmentView {
         if (mode != EnrollmentMode.NEW) {
             binding.title.visibility = View.GONE
             binding.teiDataHeader.root.visibility = View.VISIBLE
-            binding.teiDataHeader.mainAttributes.setTextColor(Color.WHITE)
-            binding.teiDataHeader.secundaryAttribute.setTextColor(Color.WHITE)
-            var firstAttr = ""
-            var secondAttr = ""
-            var thirdAttr = ""
-            if (attrList.size > 1) {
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/dhis2/usescases/enrollment/EnrollmentActivity.kt
```diff
@@ -316,29 +316,22 @@ class EnrollmentActivity : ActivityGlobalAbstract(), EnrollmentView {
...
+                    1 -> text = attrListNotEmpty[0]
+                    else -> text = String.format("%s %s", attrListNotEmpty[0], attrListNotEmpty[1])
+                }
+                setTextColor(Color.WHITE)
             }
-            if(thirdAttr.isEmpty()){
-                binding.teiDataHeader.secundaryAttribute.visibility = View.GONE
+            binding.teiDataHeader.secundaryAttribute.apply {
+                when (attrListNotEmpty.size) {
+                    0, 1, 2 -> visibility = View.GONE
```
JaimeToca: can you use constants for this magic number ?

ferdyrod: 🤣 "magic number"

JaimeToca: hahahaha I know it sounds funny XD 
https://en.wikipedia.org/wiki/Magic_number_(programming)

Balcan: I don't see them as magic numbers. Just the size of the list... would that fall into the definition?

JaimeToca: ok did not realize about the ".size"
Still, I think its a good practise to put any number/string in constants

PR #1297, opened 2020-03-27
fix: Schedulers performance fix

Balcan:


JaimeToca:
is there any jira issue for this @Balcan ? so we can test it etc..

PR #1303, opened 2020-03-30
feat: [ANDROAPP-2861,ANDROAPP-2862] Update SDK to support SMS in tracker

Balcan:
[ANDROAPP-2861](https://jira.dhis2.org/browse/ANDROAPP-2861)
[ANDROAPP-2862](https://jira.dhis2.org/browse/ANDROAPP-2862)

I tried to add some unit tests, but the SMSConfig class in the SDK is needed and it has no public access.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -219,6 +219,7 @@ dependencies {
 
     debugImplementation project(':dhis2-rule-engine')
     betaImplementation project(':dhis2-rule-engine')
+    releaseImplementation project(':dhis2-rule-engine')
 
     implementation project(':viewpagerdotsindicator')
     implementation project(':tableview')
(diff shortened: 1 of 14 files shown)


app/build.gradle
```diff
@@ -219,6 +219,7 @@ dependencies {
 
     debugImplementation project(':dhis2-rule-engine')
     betaImplementation project(':dhis2-rule-engine')
+    releaseImplementation project(':dhis2-rule-engine')
```
ferdyrod: This should not be uploaded, right?

app/build.gradle
```diff
@@ -243,9 +244,9 @@ dependencies {
         exclude group: "org.hisp.dhis", module: "core-rules"
     }
 
-    releaseImplementation("org.hisp.dhis.rules:rule-engine:${libraries.ruleEngine}") {
+  /*  releaseImplementation("org.hisp.dhis.rules:rule-engine:${libraries.ruleEngine}") {
         exclude group: "junit", module: "junit"
-    }
+    }*/
```
ferdyrod: This should not be uploaded, right?

Balcan: once we have the release version of the sdk it will revert back to default

PR #1317, opened 2020-04-02
fix: [ANDROAPP-2858] Fixes the search displayFrontPage and minimum required attribute

ferdyrod:
**JIRA ISSUE**: [ANDROAPP-2858](https://jira.dhis2.org/browse/ANDROAPP-2858)

### DIFF
--- a/app/src/main/java/org/dhis2/usescases/searchTrackEntity/SearchTEPresenter.java
+++ b/app/src/main/java/org/dhis2/usescases/searchTrackEntity/SearchTEPresenter.java
@@ -425,6 +425,7 @@ public class SearchTEPresenter implements SearchTEContractsModule.Presenter {
         view.clearList(programSelected == null ? null : programSelected.uid());
         view.clearData();
         view.setFabIcon(true);
+        showList = true;
 
         if (selectedProgram == null)
             getTrackedEntityTypeAttributes();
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/dhis2/usescases/searchTrackEntity/SearchTEPresenter.java
```diff
@@ -471,20 +472,33 @@ public void onFabClick(boolean needsSearch) {
...
+    }
+
+    private boolean compliesWithMinAttributesToSearch() {
+        if(selectedProgram != null) {
+            if(selectedProgram.displayFrontPageList() && queryData.size() < selectedProgram.minAttributesRequiredToSearch()){
+                return false;
+            } else if (!selectedProgram.displayFrontPageList() && queryData.size() < selectedProgram.minAttributesRequiredToSearch()) {
+                return false;
+            } else if (!selectedProgram.displayFrontPageList() && queryData.size() == 0 && selectedProgram.minAttributesRequiredToSearch() == 0) {
+                return false;
```
JaimeToca, line 496: do you think we could extract this and perform an unit test?

Balcan: This is already handled in 2.1.0 in a different way. I think it is not needed to add now a unit test that won't be used in the latest version

PR #1328, opened 2020-04-08
 fix(Enrollment): [ANDROAPP-2867] orgUnit not editable when is only one

mmmateos:
[Jira issue](https://jira.dhis2.org/browse/ANDROAPP-2867)

### DIFF
--- a/app/src/main/java/org/dhis2/usescases/datasets/dataSetTable/dataSetSection/DataSetSectionFragment.java
+++ b/app/src/main/java/org/dhis2/usescases/datasets/dataSetTable/dataSetSection/DataSetSectionFragment.java
@@ -1,6 +1,7 @@
 package org.dhis2.usescases.datasets.dataSetTable.dataSetSection;
 
 import android.content.Context;
+import android.content.res.Configuration;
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/dhis2/usescases/datasets/dataSetTable/dataSetSection/DataSetSectionFragment.java
```diff
@@ -81,6 +82,13 @@ public static DataSetSectionFragment create(@NonNull String sectionUid, boolean
         return dataSetSectionFragment;
     }
 
+    @Override
+    public void onConfigurationChanged(@NonNull Configuration newConfig) {
+        super.onConfigurationChanged(newConfig);
+        getAbstractActivity().hideKeyboard();
+        requireView().clearFocus();
+    }
+
```
Balcan, line 91: @mmmateos These changes are from a previous PR, don't they?

mmmateos: Yes

PR #1343, opened 2020-04-15
fix: [ANDROAPP-2901] set coordinateView when request permission

mmmateos:
[Jira Issue](https://jira.dhis2.org/browse/ANDROAPP-2901)


app/src/main/java/org/dhis2/usescases/general/ActivityGlobalAbstract.java
```diff
@@ -85,6 +85,13 @@
     public AnalyticsHelper analyticsHelper;
     public ScanTextView scanTextView;
 
+    public void requestLocationPermission(CoordinatesView coordinatesView) {
+        this.coordinatesView = coordinatesView;
+        ActivityCompat.requestPermissions((ActivityGlobalAbstract) getContext(),
+                new String[]{Manifest.permission.ACCESS_FINE_LOCATION},
+                ACCESS_LOCATION_PERMISSION_REQUEST);
+    }
```
JaimeToca, line 93: Hello I have some questions :+1: 

- Activity Global is abstract, why pass coordinates view and not "Activity"?
- is coordinates view being used?
- The purpose is to move all request permissions to "ActivityGlobalAbstract" ? or only coordinates? It's seems strange to me to only have coordinates permissions in ActivityGlobalAbstract and others in different classes.

For the future: I think all permissions could be in an external entity, class or extension outside. 

For now , I'll make the request in coordinates view and use the method "onRequestPermissionsResult" from abstract activity if it's needed...

Thoughts?

JaimeToca: ok I saw coordinates view if being used for onRequest

JaimeToca: anyway for the future. This need to be rethink.

JaimeToca: @Balcan @ferdyrod

Balcan: agreed

PR #1345, opened 2020-04-16
fix: [ANDROAPP-2904] Fixes QR Code Scan for the first time

ferdyrod:
**JIRA ISSUE**: [ANDROAPP-2904](https://jira.dhis2.org/browse/ANDROAPP-2904)


app/src/main/java/org/dhis2/usescases/login/LoginPresenter.kt
```diff
@@ -77,6 +77,8 @@ class LoginPresenter(
                                 if (!serverUrl.isNullOrEmpty() && !user.isNullOrEmpty()) {
                                     view.setUrl(serverUrl)
                                     view.setUser(user)
+                                } else {
+                                    view.setUrl(view.getDefaultServerProtocol())
```
JaimeToca, line 81: unit test :+1:  xD ?

PR #1349, opened 2020-04-17
fix: [ANDROAPP-2908] Separates Scan code reader depending on its ValueTypeRenderingType

ferdyrod:
**JIRA ISSUE**: [ANDROAPP-2908](https://jira.dhis2.org/browse/ANDROAPP-2908)


ferdyrod:
> The ScanActivity can be accessed from the LoginActivity. Have you check it still work if now it is checking the rendering type?

You are right, will need to take a look at this

PR #1350, opened 2020-04-17
fix: removed kujaku from build gradle

Balcan:


build.gradle
```diff
@@ -39,9 +39,9 @@ allprojects {
         maven {
             url 'https://maven.google.com'
         }
-        maven {
+        /*maven {
```
JaimeToca, line 42: remove the comment?

PR #1352, opened 2020-04-20
fix: [ANDROAPP-2914] Looping rules message displayed when hiding attribute from event

Balcan:
[issue](https://jira.dhis2.org/browse/ANDROAPP-2914)


JaimeToca:
@Balcan I think we have a valueStore test. It would be nice to add this small use case :)

PR #1354, opened 2020-04-20
fix: [ANDROAPP-2916] Last section padding/shadow fix

Balcan:
[issue](https://jira.dhis2.org/browse/ANDROAPP-2916)

### DIFF
--- a/app/src/main/java/org/dhis2/Bindings/Extensions.kt
+++ b/app/src/main/java/org/dhis2/Bindings/Extensions.kt
@@ -62,6 +62,7 @@ fun AppCompatActivity.app(): App {
 }
 
 val Int.dp: Int
-    get() = (this / Resources.getSystem().displayMetrics.density).toInt()
-val Int.px: Int
     get() = (this * Resources.getSystem().displayMetrics.density).toInt()
+
+val Int.px: Int
+    get() = (this / Resources.getSystem().displayMetrics.density).toInt()
--- a/app/src/main/java/org/dhis2/data/forms/dataentry/fields/section/SectionHolder.kt
+++ b/app/src/main/java/org/dhis2/data/forms/dataentry/fields/section/SectionHolder.kt
@@ -149,7 +149,7 @@ class SectionHolder(
     fun setLastSectionHeight(previousSectionIsOpened: Boolean) {
         val params = formBinding.lastSectionDetails.layoutParams
         val finalHeight = if (previousSectionIsOpened) {
-            48.px
+            48.dp
         } else {
             1.dp
         }
--- a/app/src/main/java/org/dhis2/data/forms/dataentry/fields/section/SectionHolder.kt
+++ b/app/src/main/java/org/dhis2/data/forms/dataentry/fields/section/SectionHolder.kt
@@ -1,6 +1,6 @@
 package org.dhis2.data.forms.dataentry.fields.section
 
-import android.animation.Animator
+import android.animation.ValueAnimator
 import android.view.View
 import androidx.core.content.ContextCompat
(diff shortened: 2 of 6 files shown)


app/src/main/java/org/dhis2/data/forms/dataentry/fields/section/SectionHolder.kt
```diff
@@ -138,4 +145,21 @@ class SectionHolder(
     fun setBottomShadow(showShadow: Boolean) {
         formBinding.shadowBottom.visibility = if (showShadow) View.VISIBLE else View.GONE
     }
+
+    fun setLastSectionHeight(previousSectionIsOpened: Boolean) {
+        val params = formBinding.lastSectionDetails.layoutParams
+        val finalHeight = if (previousSectionIsOpened) {
+            48.px
```
JaimeToca, line 152: can we use dp instead of px?

Balcan: It is using dp. The extension function calculates the value

JaimeToca: shouldn't it be 48.dp instead?

ferdyrod: I agree with @JaimeToca this tells me its 48 px. 😕

JaimeToca: I think the name is a bit confusing, talking to @Balcan  and it meant fromDpToPixel

Balcan: changed

app/src/main/java/org/dhis2/Bindings/Extensions.kt
```diff
@@ -62,6 +62,7 @@ fun AppCompatActivity.app(): App {
 }
 
 val Int.dp: Int
-    get() = (this / Resources.getSystem().displayMetrics.density).toInt()
+    get() = (this * Resources.getSystem().displayMetrics.density).toInt()
+
 val Int.px: Int
-    get() = (this * Resources.getSystem().displayMetrics.density).toInt()
\ No newline at end of file
+    get() = (this / Resources.getSystem().displayMetrics.density).toInt()
```
JaimeToca: Nice to have. I'd make a bit more clear.

Int.FromPixelToDP / Int.FromDpToPixel

Balcan: No, no. I changed it so that if you do 23.dp thats means 23 dps. And if you do 23.px is 23 pixels. Which I think it is way better and clearer than fromTo methods

app/src/main/java/org/dhis2/data/forms/dataentry/fields/section/SectionHolder.kt
```diff
@@ -138,4 +145,21 @@ class SectionHolder(
...
+        } else {
+            1.dp
+        }
+        ValueAnimator.ofInt(params.height, finalHeight).apply {
+            duration = 120
+            addUpdateListener {
+                params.height = it.animatedValue as Int
+                formBinding.lastSectionDetails.layoutParams = params
+            }
+            start()
```
JaimeToca, line 162: this animation should apply to all cases or only if previous section is Opened?

Balcan: All cases, open or closed.

PR #1355, opened 2020-04-20
fix: [ANDROAPP-2913] update fields after reopen event

mmmateos:
[Jira Issue](https://jira.dhis2.org/browse/ANDROAPP-2913)

### DIFF
--- a/app/src/main/java/org/dhis2/usescases/eventsWithoutRegistration/eventCapture/EventCaptureRepositoryImpl.java
+++ b/app/src/main/java/org/dhis2/usescases/eventsWithoutRegistration/eventCapture/EventCaptureRepositoryImpl.java
@@ -81,7 +81,7 @@ public class EventCaptureRepositoryImpl implements EventCaptureContract.EventCap
 
     private final FormRepository formRepository;
     private final D2 d2;
-    private final boolean isEventEditable;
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/dhis2/usescases/eventsWithoutRegistration/eventCapture/EventCaptureRepositoryImpl.java
```diff
@@ -337,7 +333,7 @@ private ProgramStageSectionRenderingType renderingType(String sectionUid) {
                             }
                         }
                         boolean editable = fieldViewModel.editable() != null ? fieldViewModel.editable() : true;
-                        fieldViewModel = fieldViewModel.withValue(value).withEditMode(editable);
+                        fieldViewModel = fieldViewModel.withValue(value).withEditMode(editable || isEventEditable);
```
Balcan, line 336: I won't use an OR operation. If the field is non editable, it should stay like that regardless of the events status. It should be an AND.

mmmateos: Okey, I'll check it

PR #1356, opened 2020-04-21
fix: [ANDROAPP-2921] Pin dialog shoud close session if clicked on forgot code

Balcan:
[issue](https://jira.dhis2.org/browse/ANDROAPP-2921)

### DIFF
--- a/app/src/main/java/org/dhis2/utils/session/PinPresenter.kt
+++ b/app/src/main/java/org/dhis2/utils/session/PinPresenter.kt
@@ -4,6 +4,7 @@ import org.dhis2.data.prefs.Preference
 import org.dhis2.data.prefs.PreferenceProvider
 import org.hisp.dhis.android.core.D2
 import org.hisp.dhis.android.core.D2Manager
+import timber.log.Timber
 import java.lang.Exception
 
 class PinPresenter(
@@ -32,7 +33,7 @@ class PinPresenter(
             preferenceProvider.setValue(Preference.PIN, null)
             preferenceProvider.setValue(Preference.SESSION_LOCKED, false)
         } catch (e: Exception) {
-
+            Timber.e(e)
         }
     }
 }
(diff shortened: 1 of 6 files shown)


app/src/main/java/org/dhis2/utils/session/PinPresenter.kt
```diff
@@ -21,4 +25,14 @@ class PinPresenter(
...
         preferenceProvider.setValue(Preference.SESSION_LOCKED, true)
     }
+
+    fun logOut() {
+        try {
+            d2?.userModule()?.blockingLogOut()
+            preferenceProvider.setValue(Preference.PIN, null)
+            preferenceProvider.setValue(Preference.SESSION_LOCKED, false)
+        } catch (e: Exception) {
+
```
JaimeToca, line 35: can you log error?

app/src/main/java/org/dhis2/utils/session/PinPresenter.kt
```diff
@@ -2,10 +2,14 @@ package org.dhis2.utils.session
...
 import org.dhis2.data.prefs.PreferenceProvider
+import org.hisp.dhis.android.core.D2
+import org.hisp.dhis.android.core.D2Manager
+import java.lang.Exception
 
 class PinPresenter(
     val view: PinView,
-    val preferenceProvider: PreferenceProvider
+    val preferenceProvider: PreferenceProvider,
+    val d2: D2? = D2Manager.getD2()
```
JaimeToca, line 12: for me it's okay to inject d2 doing it this way. But since we're using dagger to maintain consistency, could we use Pin module instead?

Balcan: The Pin module is outside the user session, so it is not possible to inject d2. I think we can refactor this once we review the full pin logic. What do you think?

app/src/main/java/org/dhis2/utils/session/PinPresenter.kt
Balcan: yes!

PR #1358, opened 2020-04-21
fix: [ANDROAPP-2922] Show assign to me filter only if a program stage is configured

Balcan:
[issue](https://jira.dhis2.org/browse/ANDROAPP-2922)

### DIFF
--- a/app/src/main/java/org/dhis2/usescases/searchTrackEntity/SearchTEActivity.java
+++ b/app/src/main/java/org/dhis2/usescases/searchTrackEntity/SearchTEActivity.java
@@ -519,10 +519,22 @@ public class SearchTEActivity extends ActivityGlobalAbstract implements SearchTE
                     Program selectedProgram = (Program) adapterView.getItemAtPosition(pos - 1);
                     setProgramColor(presenter.getProgramColor(selectedProgram.uid()));
                     presenter.setProgram((Program) adapterView.getItemAtPosition(pos - 1));
+                    if(presenter.programHasAssignment(selectedProgram.uid())){
+                        filtersAdapter.addAssignedToMe();
+                    }else{
+                        filtersAdapter.removeAssignedToMe();
+                    }
                 } else if (programs.size() == 1 && pos != 0) {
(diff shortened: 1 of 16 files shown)


app/src/main/java/org/dhis2/usescases/searchTrackEntity/SearchTEPresenter.java
```diff
@@ -810,6 +810,14 @@ public int getEnrollmentColor() {
             return -1;
     }
 
+    @Override
+    public boolean programHasAssignment(String programUid) {
+        return !d2.programModule().programStages()
+                .byProgramUid().eq(programUid)
+                .byEnableUserAssignment().isTrue()
+                .blockingIsEmpty();
+    }
```
JaimeToca, line 819: can we check here if its true or false and then do 

view.addAsignedToMe

or view.removeAssignedToMe

For the future we could unit test this logic

app/src/main/java/org/dhis2/usescases/searchTrackEntity/SearchTEActivity.java
```diff
@@ -519,10 +519,22 @@ public void onItemSelected(AdapterView<?> adapterView, View view, int pos, long
...
+                    }
                 } else if (programs.size() == 1 && pos != 0) {
                     presenter.setProgram(programs.get(0));
-                } else
+                    if(presenter.programHasAssignment(programs.get(0).uid())){
+                        filtersAdapter.addAssignedToMe();
+                    }else{
+                        filtersAdapter.removeAssignedToMe();
+                    }
+                } else {
```
JaimeToca, line 534: this could be reduce with the comment below. let me know your thoughts !

app/src/main/java/org/dhis2/usescases/searchTrackEntity/SearchTEPresenter.java
JaimeToca: and we remove logic and responsability from the view. Since it's the presenter the one to decide to show or hide... ?

PR #1364, opened 2020-04-24
docs: Update bookinfo information with updated dates [skip ci]

jaboto:
Updated dates and information on the bookinfo so the generated files match the other DHIS2 documents but with updated information.


jaboto:
Auto approving my PR as nothing critical is touched.

PR #1366, opened 2020-04-28
fix: [ANDROAPP-2985] Crash when starting app

Balcan:
[issue](https://jira.dhis2.org/browse/ANDROAPP-2985)

### DIFF
--- a/app/src/main/java/org/dhis2/usescases/login/LoginPresenter.kt
+++ b/app/src/main/java/org/dhis2/usescases/login/LoginPresenter.kt
@@ -7,7 +7,6 @@ import co.infinum.goldfinger.Goldfinger
 import io.reactivex.Observable
 import io.reactivex.disposables.CompositeDisposable
 import org.dhis2.App
-import org.dhis2.R
 import org.dhis2.data.fingerprint.FingerPrintController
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/dhis2/usescases/login/LoginPresenter.kt
```diff
@@ -253,15 +252,16 @@ class LoginPresenter(
...
         Timber.e(throwable)
         if (throwable is D2Error && throwable.errorCode() == D2ErrorCode.ALREADY_AUTHENTICATED) {
-            preferenceProvider.apply {
-                setValue(SESSION_LOCKED, false)
-
-                setValue(PIN, null)
-            }
-            view.alreadyAuthenticated()
+            userManager?.d2?.userModule()?.blockingLogOut()
+            logIn(serverUrl, userName, pass)
```
JaimeToca, line 264: do you think we can unit test this ? :+1:

Balcan: It is not that easy as the test goes into a loop... Could be done with too many changes in the presenter code for a fix.
@mmmateos , @JaimeToca , @ferdyrod Maybe you can have a try?

PR #1367, opened 2020-04-28
test(ProgramStageSelectionRepository): migrates to kotlin ands add test

mmmateos:


app/src/main/java/org/dhis2/usescases/programStageSelection/ProgramStageSelectionRepositoryImpl.kt
```diff
@@ -0,0 +1,144 @@
...
+    private val programUid: String,
+    private val enrollmentUid: String,
+    private val eventCreationType: String,
+    private val d2: D2
+) : ProgramStageSelectionRepository {
+
+    private val cachedRuleEngineFlowable: Flowable<RuleEngine>
+
+    @VisibleForTesting
+    fun ruleEnrollment(): Flowable<RuleEnrollment> =
```
ferdyrod, line 40: All methods in Kotlin are public by default. There is no need for the annotation in here. 👍

PR #1369, opened 2020-04-29
fix(EnrollmentActivity): [ANDROAPP-2982] check uuid is not null

mmmateos:
[Jira Issue](https://jira.dhis2.org/browse/ANDROAPP-2982)


JaimeToca:
Hello, this fix prevents the app from crashing but does not solve the problem ? Or maybe yes..  but it would be nice to track why this is actually happening. Do you think you could track a non-fatal exception in firebase. Thoughts?

PR #1371, opened 2020-04-29
fix: [ANDROAPP-2983] Fix crash on ProgramEventDetailRepository

ferdyrod:
**JIRA ISSUE**: [ANDROAPP-2983](https://jira.dhis2.org/browse/ANDROAPP-2983)

### DIFF
--- a/app/src/main/java/org/dhis2/usescases/programEventDetail/ProgramEventMapper.kt
+++ b/app/src/main/java/org/dhis2/usescases/programEventDetail/ProgramEventMapper.kt
@@ -1,0 +1,154 @@
+package org.dhis2.usescases.programEventDetail
+
+import org.dhis2.Bindings.userFriendlyValue
+import org.dhis2.data.tuples.Pair
+import org.dhis2.utils.DateUtils
+import org.hisp.dhis.android.core.D2
+import org.hisp.dhis.android.core.arch.helpers.UidsHelper
+import org.hisp.dhis.android.core.arch.repositories.scope.RepositoryScope
+import org.hisp.dhis.android.core.category.CategoryOptionCombo
+import org.hisp.dhis.android.core.common.State
+import org.hisp.dhis.android.core.event.Event
+import org.hisp.dhis.android.core.program.ProgramStageSection
+import org.hisp.dhis.android.core.trackedentity.TrackedEntityDataValue
+import java.util.Collections
+import java.util.Date
+import javax.inject.Inject
+
+class ProgramEventMapper @Inject constructor(val d2: D2) {
+
+    fun eventToProgramEvent(event: Event): ProgramEventViewModel {
+        val orgUnitName: String = getOrgUnitName(event.organisationUnit()) ?: ""
+        val showInReportsDataElements = mutableListOf<String>()
+        val programStageDataElements = getProgramStageDataElements(event.programStage())
+        programStageDataElements.forEach {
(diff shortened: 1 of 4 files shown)


app/src/main/java/org/dhis2/usescases/programEventDetail/ProgramEventMapper.kt
```diff
@@ -0,0 +1,154 @@
...
+        val orgUnitName: String = getOrgUnitName(event.organisationUnit()) ?: ""
+        val showInReportsDataElements = mutableListOf<String>()
+        val programStageDataElements = getProgramStageDataElements(event.programStage())
+        programStageDataElements.forEach {
+            if(it.displayInReports() == true) {
+                it.dataElement()?.let { dataElement ->
+                    showInReportsDataElements.add(dataElement.uid())
+                }
+            }
+        }
```
JaimeToca, line 30: I think you can use filter instead :+1:

app/src/main/java/org/dhis2/usescases/programEventDetail/ProgramEventMapper.kt
```diff
@@ -0,0 +1,154 @@
...
+        )
+        val hasExpired = isExpired(event)
+        val inOrgUnitRange = checkOrgUnitRange(event.organisationUnit(), event.eventDate()!!)
+        val catOptComb = getCategoryOptionCombo(event.attributeOptionCombo())
+        val attrOptCombo: String =
+            if (catOptComb != null && !catOptComb.displayName().equals("default")) {
+                catOptComb.displayName()!!
+            } else {
+                ""
+            }
```
JaimeToca, line 44: I think you can use ? and remove !! that throws exceptions. Also in your "if" in case displayName is null it can crash

if (!catOptComb?.displayName()?.equals("default")) {
                catOptComb.displayName()
            } else {
                ""
            }

or ternary operator

app/src/main/java/org/dhis2/usescases/programEventDetail/ProgramEventMapper.kt
```diff
@@ -0,0 +1,154 @@
...
+        }
+        val data = getData(
+            event.trackedEntityDataValues(),
+            showInReportsDataElements,
+            event.programStage()
+        )
+        val hasExpired = isExpired(event)
+        val inOrgUnitRange = checkOrgUnitRange(event.organisationUnit(), event.eventDate()!!)
+        val catOptComb = getCategoryOptionCombo(event.attributeOptionCombo())
+        val attrOptCombo: String =
```
JaimeToca, line 39: getData, getProgram, getOrgUnitName, getCategoryOptionCombo these methods access the database and I think their best place to be is the repository not the mapper. You could actually pass the repository to the mapper, or pass a dataclass with the information needed.

isExpired and checkOrgUnitRange is fine to be here :+1: 

Thoughts ??

Balcan: Agree. Would be better to keep all access to the database in the repository classes

ferdyrod: I agree, but will doing this it is getting a little out of hand to comply with the date to release an apk 2.1.1 bugfixed. It's better if this goes to another issue when there is more time.

PR #1372, opened 2020-04-29
fix: [ANDROAPP-2668] Support RTL in dataSet tables

Balcan:
[issue](https://jira.dhis2.org/browse/ANDROAPP-2668)


tableview/src/main/java/com/evrencoskun/tableview/TableView.java
```diff
@@ -139,6 +140,8 @@ public TableView(@NonNull Context context, @Nullable AttributeSet attrs, @AttrRe
     }
 
     private void initialDefaultValues(AttributeSet attrs) {
+        //RTL
+        isRTL = getContext().getResources().getConfiguration().getLayoutDirection() == LAYOUT_DIRECTION_RTL;
```
JaimeToca, line 144: are you using this value?

Balcan: done!

PR #1382, opened 2020-05-05
fix: [ANDROAPP-3021] don't count fields without section for percentage

mmmateos:
[Jira Issue](https://jira.dhis2.org/browse/ANDROAPP-3021)

### DIFF
--- a/app/src/main/java/org/dhis2/usescases/eventsWithoutRegistration/eventCapture/EventCapturePresenterImpl.java
+++ b/app/src/main/java/org/dhis2/usescases/eventsWithoutRegistration/eventCapture/EventCapturePresenterImpl.java
@@ -221,19 +221,21 @@ public class EventCapturePresenterImpl implements EventCaptureContract.Presenter
                                                     List<String> optionSets = new ArrayList<>();
                                                     for (FieldViewModel fieldViewModel : fields) {
                                                         String fieldSection = getFieldSection(fieldViewModel);
-                                                        if (!fieldMap.containsKey(fieldSection)) {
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/dhis2/usescases/eventsWithoutRegistration/eventCapture/EventCapturePresenterImpl.java
```diff
@@ -221,19 +221,21 @@ public void init() {
...
-                                                            optionSets.add(fieldViewModel.optionSet());
-                                                            totalFields++;
+                                                            if (fieldViewModel.optionSet() == null || !(fieldViewModel instanceof ImageViewModel)) {
+                                                                totalFields++;
+                                                            } else if (!optionSets.contains(fieldViewModel.optionSet())) {
+                                                                optionSets.add(fieldViewModel.optionSet());
+                                                                totalFields++;
+                                                            }
+                                                            if (fieldViewModel instanceof UnsupportedViewModel || fieldViewModel instanceof DisplayViewModel)
+                                                                unsupportedFields++;
```
JaimeToca, line 237: it's okay, for next time @mmmateos  @ferdyrod  @Balcan  I'll push for having at least some unit testing regarding init method. It does not need to perform everything but  sectionList can easily be mocked and fieldMap asserted

JaimeToca: Remember, for future. If we don't have unit test. We'll have to add espresso which are expensive tests

PR #1384, opened 2020-05-05
fix: [ANDROAPP-3024] Do not hide filters icon if no data is returned

Balcan:
[issue](https://jira.dhis2.org/browse/ANDROAPP-3024)

### DIFF
--- a/app/src/main/java/org/dhis2/usescases/searchTrackEntity/SearchTEActivity.java
+++ b/app/src/main/java/org/dhis2/usescases/searchTrackEntity/SearchTEActivity.java
@@ -444,19 +444,13 @@ public class SearchTEActivity extends ActivityGlobalAbstract implements SearchTE
         if (!fromRelationship) {
             liveData.observe(this, searchTeiModels -> {
                 Trio<PagedList<SearchTeiModel>, String, Boolean> data = presenter.getMessage(searchTeiModels);
+                presenter.checkFilters(data.val1().isEmpty());
                 if (data.val1().isEmpty()) {
-                    binding.filterCounter.setVisibility(View.VISIBLE);
-                    binding.searchFilterGeneral.setVisibility(View.VISIBLE);
-
                     binding.messageContainer.setVisibility(View.GONE);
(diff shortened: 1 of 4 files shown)


app/src/main/java/org/dhis2/usescases/searchTrackEntity/SearchTEActivity.java
```diff
@@ -453,8 +453,9 @@ public void setLiveData(LiveData<PagedList<SearchTeiModel>> liveData) {
                     liveAdapter.submitList(data.val0());
                     binding.progressLayout.setVisibility(View.GONE);
                 } else {
-                    binding.filterCounter.setVisibility(View.GONE);
-                    binding.searchFilterGeneral.setVisibility(View.GONE);
+                    boolean filtersActive = FilterManager.getInstance().getTotalFilters() != 0;
+                    binding.filterCounter.setVisibility(filtersActive ? View.VISIBLE : View.GONE);
+                    binding.searchFilterGeneral.setVisibility(filtersActive ? View.VISIBLE : View.GONE);
```
JaimeToca, line 458: I know that this class is not new, but what do you think if we do this instead 

presenter.updateFilterCount(filtersActive)

SearchTePresenter.class

fun updateFilterCount(isActive:Boolean){
   if (isActive) do something
   else do this
}

Then it can be easily add to our searchTepresenterTest and reduce this debt

Let me know what you think about it

Balcan: The filterCounter already updates the value using an observableField. This is just to not hide the filter button which requires first to have the result from the search.

JaimeToca: Sorry I meant FilterCounter XD

fun updateFilterCounter ...

Balcan: done

PR #1385, opened 2020-05-05
fix: [ANDROAPP-3022] Fixes bug with back button on enrollment and event forms

ferdyrod:
**JIRA ISSUE**: [ANDROAPP-3022](https://jira.dhis2.org/browse/ANDROAPP-3022)

### DIFF
--- a/app/src/main/java/org/dhis2/Bindings/Extensions.kt
+++ b/app/src/main/java/org/dhis2/Bindings/Extensions.kt
@@ -1,6 +1,8 @@
 package org.dhis2.Bindings
 
 import android.content.res.Resources
+import android.graphics.Rect
+import android.view.ViewGroup
 import androidx.appcompat.app.AppCompatActivity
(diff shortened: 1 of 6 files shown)


app/src/main/java/org/dhis2/Bindings/Extensions.kt
```diff
@@ -66,3 +68,17 @@ val Int.dp: Int
 
 val Int.px: Int
     get() = (this / Resources.getSystem().displayMetrics.density).toInt()
+
+fun AppCompatActivity.isKeyboardOpened(): Boolean {
+    val r = Rect()
+    val keyboardVisibilityThreshold = 100
+
+    val activityRoot = (findViewById<ViewGroup>(android.R.id.content)).getChildAt(0)
+    val visibleThreshold = keyboardVisibilityThreshold.dp
```
JaimeToca, line 77: if (activityRootView == null) return false; ?

PR #1502, opened 2020-07-13
chore: [ANDROAPP-3174] Changes Events fragments title

ferdyrod:
## Description

[ANDROAPP-3174](https://jira.dhis2.org/browse/ANDROAPP-3174)

- [ ] Smartphone Emulator
- [ ] Tablet Emulator
- [x] Smartphone
- [ ] Tablet
- [ ] 4.4
- [ ] 5.X - 6.X
- [ ] 7.X
- [ ] 8.X
- [x] 9.X - 10.X
- [ ] Other
## Checklist


Balcan:
@ferdyrod I approved this PR but I think we should have a new string resource like **event_data** so that the translation platforms knows something new has been added and users can translate this new text. If we only change the text the users won't be notified.

ferdyrod:
> @ferdyrod I approved this PR but I think we should have a new string resource like **event_data** so that the translation platforms knows something new has been added and users can translate this new text. If we only change the text the users won't be notified.

I asked about this in the slack channel, got no response! 😢
I'll add a new string resource.

PR #1534, opened 2020-07-24
docs: Fix comment about disabled account [skip ci]

jaboto:


jaboto:
I have added some minor grammar fixes, the SMS doc is not yet indexed so approving the PR has no impact until the INDEX file is modified.

PR #1537, opened 2020-07-27
fix: [ANDROAPP-3167] truncates coordinates values when entered manually by user.

ferdyrod:
## Description

[ANDROAPP-3167](https://jira.dhis2.org/browse/ANDROAPP-3167)

## Solution description
## Covered unit test cases
- [x] Smartphone Emulator
- [ ] Tablet Emulator
- [x] Smartphone
- [ ] Tablet
- [ ] 4.4
- [ ] 5.X - 6.X
- [ ] 7.X
- [ ] 8.X
- [x] 9.X - 10.X
- [ ] Other
## Checklist

### DIFF
--- a/app/src/main/java/org/dhis2/Bindings/StringExtensions.kt
+++ b/app/src/main/java/org/dhis2/Bindings/StringExtensions.kt
@@ -64,3 +64,5 @@ fun String.toDate(): Date {
 
     return date
 }
+
+fun String.parseToDouble() = this.toDoubleOrNull()?.toString() ?: "0.0"
--- a/app/src/main/java/org/dhis2/usescases/coodinates/CoordinatesView.java
+++ b/app/src/main/java/org/dhis2/usescases/coodinates/CoordinatesView.java
@@ -148,6 +148,28 @@ public class CoordinatesView extends FieldLayout implements View.OnClickListener
         position.setOnClickListener(this);
         map.setOnClickListener(this);
         clearButton.setOnClickListener(this);
+
+        longitude.setOnFocusChangeListener(new OnFocusChangeListener() {
+            @Override
+            public void onFocusChange(View v, boolean hasFocus) {
+                if(!hasFocus)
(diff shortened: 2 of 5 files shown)


app/src/main/java/org/dhis2/usescases/coodinates/CoordinatesView.java
```diff
@@ -148,6 +148,28 @@ private void setLayout() {
...
         map.setOnClickListener(this);
         clearButton.setOnClickListener(this);
+
+        longitude.setOnFocusChangeListener(new OnFocusChangeListener() {
+            @Override
+            public void onFocusChange(View v, boolean hasFocus) {
+                if(!hasFocus)
+                    if (!longitude.getText().toString().isEmpty()) {
+                        Double lon = DoubleExtensionsKt.truncate(getLongitude());
+                        longitude.setText(lon.toString());
```
JaimeToca, line 158: I think you need to have a regex or validate the entry. Otherwise if I don't write the coordinates properly after changing focus this could crash?

What if I write "1." and change focus?

ferdyrod: In the first two cases a "0" is added appropriately. There wasn't any validation before but will added. The same happens with the "-" sign as the ".".

app/src/main/java/org/dhis2/Bindings/StringExtensions.kt
```diff
@@ -64,3 +64,5 @@ fun String.toDate(): Date {
 
     return date
 }
+
+fun String.parseToDouble() = this.toDoubleOrNull()?.toString() ?: "0.0"
```
JaimeToca, line 68: I think it would be a nice idea to validate somehow -90/90 and -180/180. Otherwise that could be upload to the server and when entering the dashboard get an alert dialog about wrong coordinates. What do you think?

ferdyrod: done!

PR #1544, opened 2020-07-28
fix: [ANDROAPP-3140] don't lose focus when change focus manually

mmmateos:
[ jira issue ](https://jira.dhis2.org/browse/ANDROAPP-3140)

## Solution description
We should keep field focus to be able to open the corresponding keyboard.
Also, we have to keep focus on selected item when selecting a field that is above current

- [ ] Smartphone Emulator
- [ ] Tablet Emulator
- [ ] Smartphone
- [x] Tablet
- [ ] 4.4
- [ ] 5.X - 6.X
- [x] 7.X
- [ ] 8.X
- [ ] 9.X - 10.X
- [ ] Other
## Checklist

### DIFF
--- a/app/src/main/java/org/dhis2/data/forms/dataentry/fields/FormViewHolder.java
+++ b/app/src/main/java/org/dhis2/data/forms/dataentry/fields/FormViewHolder.java
@@ -28,6 +28,7 @@ public abstract class FormViewHolder extends RecyclerView.ViewHolder {
     protected MutableLiveData<String> currentUid;
     protected String fieldUid;
     protected ObjectStyle objectStyle;
+    protected static String selectedFieldUid;
 
     public FormViewHolder(ViewDataBinding binding) {
         super(binding.getRoot());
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/dhis2/data/forms/dataentry/fields/FormViewHolder.java
```diff
@@ -50,7 +51,7 @@ public FormViewHolder(ViewDataBinding binding) {
     public void initFieldFocus() {
         if (currentUid != null) {
             currentUid.observeForever(fieldUid -> {
-                if (Objects.equals(fieldUid, this.fieldUid)) {
+                if((selectedFieldUid == null && Objects.equals(fieldUid, this.fieldUid)) || ((selectedFieldUid != null) && selectedFieldUid.equals(this.fieldUid))){
```
Balcan, line 54: Too difficult to follow, could you extract it to a method call with a friendly name?

PR #1583, opened 2020-08-13
fix: [ANDROAPP-3033] Open next section if current selections is hidden

Balcan:
## Description

[ jira issue ](https://jira.dhis2.org/browse/ANDROAPP-3033)

## Solution description
## Covered unit test cases
- [ ] Smartphone Emulator
- [ ] Tablet Emulator
- [ ] Smartphone
- [ ] Tablet
- [ ] 4.4
- [ ] 5.X - 6.X
- [ ] 7.X
- [ ] 8.X
- [ ] 9.X - 10.X
- [ ] Other
## Checklist


app/src/main/java/org/dhis2/usescases/eventsWithoutRegistration/eventCapture/EventCapturePresenterImpl.java
```diff
@@ -349,6 +352,24 @@ public void init() {
...
+                        activeSection = sectionList.get(nextSectionIndex).sectionUid();
+                    }else{
+                        activeSection = "";
+                    }
+                    break;
+                }
+            }
+        }
+        return activeSection;
+    }
```
JaimeToca, line 371: I think this could go to another class and maybe reuse in the future? The fact that you need to add @VisibleForTesting here tells you too many things are going on in the flatmap. Pls do not make an extension :+1:

app/src/main/java/org/dhis2/usescases/eventsWithoutRegistration/eventCapture/EventCapturePresenterImpl.java
```diff
@@ -349,6 +352,24 @@ public void init() {
         fieldFlowable.connect();
     }
 
+    @VisibleForTesting
+    public String getNextVisibleSection(String activeSection, List<FormSectionViewModel> sectionList) {
+        while (sectionsToHide.contains(activeSection)) {
```
JaimeToca, line 357: I think you don't need a while, "if" if enough. Right?

app/src/main/java/org/dhis2/usescases/eventsWithoutRegistration/eventCapture/EventCapturePresenterImpl.java
JaimeToca: Also it could be done in Kotlin with collection operations

JaimeToca:
Let me know your thoughts about this @Balcan 
```
class GetNextVisibleSection {

    fun get(
        activeSection: String,
        allSections: List<FormSectionViewModel>,
        sectionsToHide: List<String>
    ): String {

        if (sectionsToHide.isEmpty()) return activeSection

        val sectionsToShow = allSections.map { it.sectionUid() }.minus(sectionsToHide)

        return if (sectionsToShow.isEmpty()){
            EMPTY_SECTION
        } else {
            sectionsToShow.first()!!
        }
    }

    companion object {
        const val EMPTY_SECTION = ""
    }
}
```

mmmateos (approved):

PR #1587, opened 2020-08-13
fix: [ANDROAPP-3276] Hide schedule option if hideDueDate is true

Balcan:
## Description

[ jira issue ](https://jira.dhis2.org/browse/ANDROAPP-3276)

## Solution description
## Covered unit test cases
- [ ] Smartphone Emulator
- [ ] Tablet Emulator
- [ ] Smartphone
- [ ] Tablet
- [ ] 4.4
- [ ] 5.X - 6.X
- [ ] 7.X
- [ ] 8.X
- [ ] 9.X - 10.X
- [ ] Other
## Checklist

### DIFF
--- a/app/src/main/java/org/dhis2/usescases/teiDashboard/dashboardfragments/teidata/TEIDataFragment.java
+++ b/app/src/main/java/org/dhis2/usescases/teiDashboard/dashboardfragments/teidata/TEIDataFragment.java
@@ -522,6 +522,7 @@ public class TEIDataFragment extends FragmentGlobalAbstract implements TEIDataCo
     public void showNewEventOptions(View anchor, ProgramStage stage) {
         PopupMenu popupMenu = new PopupMenu(context, anchor);
         popupMenu.inflate(R.menu.dashboard_event_creation);
+        popupMenu.getMenu().findItem(R.id.schedulenew).setVisible(!stage.hideDueDate());
         popupMenu.setOnMenuItemClickListener(item -> {
             switch (item.getItemId()) {
                 case R.id.schedulenew:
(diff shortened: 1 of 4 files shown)


app/src/main/java/org/dhis2/usescases/teiDashboard/dashboardfragments/teidata/TEIDataFragment.java
```diff
@@ -522,6 +522,7 @@ public void showTeiImage(String filePath, String defaultIcon) {
     public void showNewEventOptions(View anchor, ProgramStage stage) {
         PopupMenu popupMenu = new PopupMenu(context, anchor);
         popupMenu.inflate(R.menu.dashboard_event_creation);
+        popupMenu.getMenu().findItem(R.id.schedulenew).setVisible(!stage.hideDueDate());
```
JaimeToca, line 525: Nice :+1: 

I would do a couple of things:

1) Add this check in the presenter inside void onAddNewEvent, and did a tiny unit test. I know is very small but we'll avoid checking this kind of things in espresso tests.

2) I know this name comes from SDK? (This is a personal preference that wanted to share) 

```
if (account.isEnabled) {
  // ...
}
It's easier to read than
if (!account.isDisabled) {
  // ...
}
```
So perhaps we could do something like:

If (stage.hideDueDate()) popupMenu.getMenu().findItem(R.id.schedulenew).setVisible(false)
//do the else with (true)

JaimeToca: added unit test. @Balcan

PR #1672, opened 2020-09-28
fix: ANDROAPP-3332 Update map when event changes coordinates

Balcan:
## Description

[ jira issue ](https://jira.dhis2.org/browse/ANDROAPP-3332)

## Solution description
## Covered unit test cases
- [ ] Smartphone Emulator
- [ ] Tablet Emulator
- [ ] Smartphone
- [ ] Tablet
- [ ] 4.4
- [ ] 5.X - 6.X
- [ ] 7.X
- [ ] 8.X
- [ ] 9.X - 10.X
- [ ] Other
## Checklist

### DIFF
--- a/app/src/main/java/org/dhis2/usescases/eventsWithoutRegistration/eventInitial/EventInitialActivity.java
+++ b/app/src/main/java/org/dhis2/usescases/eventsWithoutRegistration/eventInitial/EventInitialActivity.java
@@ -501,6 +501,7 @@ public class EventInitialActivity extends ActivityGlobalAbstract implements Even
         binding.geometry.setCurrentLocationListener(geometry -> {
             Timber.tag("EVENTINITIAL").d("NEW GEOMETRY");
             this.newGeometry = geometry;
+            preferences.setValue(Preference.EVENT_COORDINATE_CHANGED,true);
         });
         binding.geometry.setMapListener(
                 (CoordinatesView.OnMapPositionClick) binding.geometry.getContext()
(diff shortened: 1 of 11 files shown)


app/src/main/java/org/dhis2/usescases/eventsWithoutRegistration/eventInitial/EventInitialPresenter.java
```diff
@@ -238,6 +238,7 @@ private void getProgramStages(String programUid, String programStageUid) {
 
     @Override
     public void onBackClick() {
+        preferences.setValue(Preference.EVENT_COORDINATE_CHANGED, false);
```
JaimeToca, line 241: do we need to clear this values?

app/src/main/java/org/dhis2/usescases/eventsWithoutRegistration/eventInitial/EventInitialActivity.java
```diff
@@ -501,6 +501,7 @@ public void setProgramStage(ProgramStage programStage) {
         binding.geometry.setCurrentLocationListener(geometry -> {
             Timber.tag("EVENTINITIAL").d("NEW GEOMETRY");
             this.newGeometry = geometry;
+            preferences.setValue(Preference.EVENT_COORDINATE_CHANGED,true);
```
JaimeToca, line 504: In some cases we access preferences in activity (like this one) in other cases in presenter. I think we should maintaining an architecture order and access data as always from presenter, also having preferences in activity makes to the testing more difficult.

Maybe you could do this in presenter like the other save. In fact, you could have a method name.

fun name(hasCoordinateChange: Boolean){
preferences.setValue(Preference.EVENT_COORDINATE_CHANGED, hasCoordinateChange);
}

thoughts?

app/src/main/java/org/dhis2/usescases/eventsWithoutRegistration/eventInitial/EventInitialPresenter.java
Balcan: Yes. Clearing or setting it to false will do the same.

JaimeToca (approved):

mmmateos (approved):

PR #1737, opened 2020-10-19
docs: Update docs info to say 2.3 [skip ci]

jaboto:
Added info on the guidelines to explain that they are valid for 2.3


mmvila (approved):
Thanks Jaime.

PR #1740, opened 2020-10-21
fix: [ANDROAPP-3475] Sending jira issue from app is not working

Balcan:
## Description
**Note**: Candidate for 2.3.1
[ jira issue ](https://jira.dhis2.org/browse/ANDROAPP-3475)

## Solution description
## Covered unit test cases
- [ ] Smartphone Emulator
- [ ] Tablet Emulator
- [ ] Smartphone
- [ ] Tablet
- [ ] 4.4
- [ ] 5.X - 6.X
- [ ] 7.X
- [ ] 8.X
- [ ] 9.X - 10.X
- [ ] Other
## Checklist


app/src/main/java/org/dhis2/data/jira/JiraIssueListResponse.kt
```diff
@@ -0,0 +1,25 @@
...
+    val issues: List<JiraIssue>
+) {
+    companion object {
+        fun fromJson(response: String?): JiraIssueListResponse {
+            return try {
+                Gson()
+                    .fromJson(
+                        response,
+                        JiraIssueListResponse::class.java
+                    )
```
JaimeToca, line 18: Do you think is a good idea to add a ConverterFactory(GsonConverterFactory.create()) to retrofit initialization and avoid this?

app/src/main/java/org/dhis2/usescases/jira/JiraRepository.kt
```diff
@@ -0,0 +1,110 @@
...
+                        onResponse(JiraIssuesResult(issueList))
+                    } else {
+                        onResponse(JiraIssuesResult(errorMessage = response.message()))
+                    }
+                }
+
+                override fun onFailure(call: Call<ResponseBody>, t: Throwable) {
+                    onResponse(JiraIssuesResult(errorMessage = t.localizedMessage))
+                }
+            }
```
JaimeToca, line 62: What do you think about removing the enqueue callback and make it sync, then using IO RxJava? You could remove the callback. 

Thoughts?

app/src/main/java/org/dhis2/usescases/jira/JiraViewModel.kt
```diff
@@ -1,208 +1,138 @@
...
+        } else {
+            issueMessage.value = resources.jiraIssueSentErrorMessage()
+        }
     }
 
-    override fun onJiraPassChanged(s: CharSequence, start: Int, before: Int, count: Int) {
-        pass.value = s.toString()
+    @VisibleForTesting
+    fun handleThrowableResponse(throwable: Throwable) {
+        issueMessage.value = throwable.localizedMessage
```
JaimeToca, line 117: I don't think you need to use VisibleforTesting. With Mockito you can mocks callbacks and choose either success or error, then you can make the assertions. Let me know if you need help with this !

JaimeToca:
@Balcan Great Job, really nice PR

app/src/main/java/org/dhis2/data/jira/JiraIssueListResponse.kt
Balcan: yes, that would be nice 👍

app/src/main/java/org/dhis2/usescases/jira/JiraRepository.kt
Balcan: I'm going to look at this. My only concern is dealing with the disposables and the viewmodel lifecycle at the same time...

app/src/main/java/org/dhis2/usescases/jira/JiraViewModel.kt
Balcan: true

JaimeToca (approved):

PR #1836, opened 2020-12-22
fix: [ANDROAPP-3575] ScanView crashed when error was null

ferdyrod:
## Description

[[ANDROAPP-3575](https://jira.dhis2.org/browse/[ANDROAPP-3575)

## Solution description
## Covered unit test cases
- [ ] Smartphone Emulator
- [ ] Tablet Emulator
- [ ] Smartphone
- [ ] Tablet
- [ ] 4.4
- [ ] 5.X - 6.X
- [ ] 7.X
- [ ] 8.X
- [ ] 9.X - 10.X
- [ ] Other
## Checklist

### DIFF
--- a/app/src/main/java/org/dhis2/data/forms/dataentry/fields/FieldViewModel.java
+++ b/app/src/main/java/org/dhis2/data/forms/dataentry/fields/FieldViewModel.java
@@ -43,7 +43,7 @@ public abstract class FieldViewModel {
     public abstract FieldViewModel withWarning(@NonNull String warning);
 
     @NonNull
-    public abstract FieldViewModel withError(@NonNull String error);
+    public abstract FieldViewModel withError(@Nullable String error);
 
     @Nullable
     public abstract String description();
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/dhis2/data/forms/dataentry/fields/FieldViewModel.java
```diff
@@ -43,7 +43,7 @@
     public abstract FieldViewModel withWarning(@NonNull String warning);
 
     @NonNull
-    public abstract FieldViewModel withError(@NonNull String error);
+    public abstract FieldViewModel withError(@Nullable String error);
```
andresmr, line 46: do you think it make sense having a method called withError with nullable parameter?.
Maybe we can check when are we calling it with null , maybe we can avoid `ScanView` to call `"withError"` with null value.
What do you think?

andresmr (changes requested):

app/src/main/java/org/dhis2/data/forms/dataentry/fields/FieldViewModel.java
ferdyrod: Its better that way, I'm changing it.

andresmr (approved):

mmmateos (approved):

PR #1859, opened 2021-01-26
feat: [ANDROAPP-1161] display indicators in DataSets

mmmateos:
## Description

[ jira issue ](https://jira.dhis2.org/browse/ANDROAPP-1161)

## Solution description
## Covered unit test cases
- [ ] Smartphone Emulator
- [ ] Tablet Emulator
- [ ] Smartphone
- [ ] Tablet
- [ ] 4.4
- [ ] 5.X - 6.X
- [ ] 7.X
- [ ] 8.X
- [ ] 9.X - 10.X
- [ ] Other
## Checklist


app/src/main/java/org/dhis2/usescases/datasets/dataSetTable/dataSetSection/DataSetIndicatorHolder.kt
```diff
@@ -0,0 +1,34 @@
...
+import androidx.core.content.ContextCompat
+import androidx.core.view.ViewCompat
+import com.evrencoskun.tableview.adapter.recyclerview.holder.AbstractViewHolder
+import dhis2.org.R
+
+class DataSetIndicatorHolder(itemView: View) : AbstractViewHolder(itemView) {
+    fun bind(text: String, isHeader: Boolean) {
+        itemView.findViewById<TextView>(R.id.text).let {
+            it.text = text
+            if (isHeader) {
```
andresmr, line 15: I wouldn't add presentation logic to the holder, when you call this method in the _DataSetIndicatorAdapter_, either for the header (line 35) or the row (line 49) you have the data models, maybe the model could have Ui properties like `backgroundColor: Int` and `textcolor: Int`.
What do you think?

andresmr (changes requested):

app/src/main/java/org/dhis2/usescases/datasets/dataSetTable/dataSetSection/DataSetIndicatorHolder.kt
ferdyrod: I would agree with @andresmr in here, however I don't know if we have a data model in here. Looking at the methods in the adapter this "models" can be anything (Any), so don't know if we can a proper data model for this holder.

or am I wrong??

andresmr (approved):

app/src/main/java/org/dhis2/usescases/datasets/dataSetTable/dataSetSection/DataSetIndicatorHolder.kt
andresmr: Yes, you are right, they are set toString when calling binding, well I don't know is this would be out of the scope of the ticket, but we could think how these data models are created.

ferdyrod (approved):

andresmr (approved):

PR #1872, opened 2021-01-29
feat: [ANDROAPP-3312] Merge/show color by legends in event entry form

ifoche:
## Description
This PR implements the functionality of adding colors to the event form, by using the legends downloaded by the SDK. It depends on the code also submitted in the following PR: https://github.com/dhis2/dhis2-android-sdk/pull/1469


app/src/main/java/org/dhis2/data/forms/dataentry/fields/edittext/EditTextCustomHolder.java
```diff
@@ -135,6 +141,16 @@ public void update(@NonNull FieldViewModel model) {
         setLongClick();
     }
 
+    private void assignBackgroundColorByLegend() {
+        if (editTextModel.colorByLegend() != null && editTextModel.colorByLegend() != ""){
```
aimiebell, line 145: I detect that this code is problematic. According to the [Bad practice (BAD_PRACTICE)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#bad-practice-bad-practice), [ES: Comparison of String objects using == or != (ES_COMPARING_STRINGS_WITH_EQ)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#es-comparison-of-string-objects-using-or-es-comparing-strings-with-eq).
This code compares java.lang.String objects for reference equality using the == or != operators. Unless both strings are either constants in a source file, or have been interned using the String.intern() method, the same string value may be represented by two different String objects. Consider using the equals(Object) method instead.

app/src/main/java/org/dhis2/data/forms/dataentry/fields/spinner/SpinnerHolder.java
```diff
@@ -57,6 +60,19 @@ public void update(SpinnerViewModel viewModel) {
...
         label = new StringBuilder().append(viewModel.label());
         initFieldFocus();
+
+        if(!isSearchMode){
+            assignBackgroundColorByLegend();
+        }
+    }
+
+    private void assignBackgroundColorByLegend() {
+        if (viewModel.colorByLegend() != null && viewModel.colorByLegend() != ""){
```
aimiebell, line 70: I detect that this code is problematic. According to the [Bad practice (BAD_PRACTICE)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#bad-practice-bad-practice), [ES: Comparison of String objects using == or != (ES_COMPARING_STRINGS_WITH_EQ)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#es-comparison-of-string-objects-using-or-es-comparing-strings-with-eq).
This code compares java.lang.String objects for reference equality using the == or != operators. Unless both strings are either constants in a source file, or have been interned using the String.intern() method, the same string value may be represented by two different String objects. Consider using the equals(Object) method instead.

Balcan:
@ifoche @xurxodev 
I'm closing this PR in favor of [this ine](https://github.com/dhis2/dhis2-android-capture-app/pull/1913). Thanks for the contribution! :)

PR #1964, opened 2021-03-11
feat: [ANDROAPP-3760] Add matomo trackers

JaimeToca:
## Description

## Solution description
## Covered unit test cases
- [ ] Smartphone Emulator
- [ ] Tablet Emulator
- [ ] Smartphone
- [ ] Tablet
- [ ] 4.4
- [ ] 5.X - 6.X
- [ ] 7.X
- [ ] 8.X
- [ ] 9.X - 10.X
- [ ] Other
## Checklist

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -55,7 +55,7 @@ android {
         buildConfigField "String", "SDK_VERSION", "\"" + libraries.dhis2sdk + "\""
         buildConfigField "String", "MAPBOX_ACCESS_TOKEN", "\"pk.eyJ1Ijoib3Jpa2dnIiwiYSI6ImNqdWpqb24zYTFsOXo0M2xwcXRseGx3aXkifQ.6N9aqvpqdrrSBwO9GyfjUw\""
         buildConfigField "String", "ACRA_URL", "\"$acraUrl\""
-        buildConfigField "String", "MATOMO_URL", "\"https://127.0.0.1:8080/matomo.php\""
+        buildConfigField "String", "MATOMO_URL", "\"https://usage.analytics.dhis2.org/matomo.php\""
 
         javaCompileOptions {
             annotationProcessorOptions {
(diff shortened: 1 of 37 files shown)


Balcan (approved):
My only comment. Wouldn't it be better if we move all tracking constants to a single class file? This way we can avoid duplications or name changes between screens.

ferdyrod (approved):

JaimeToca:
> My only comment. Wouldn't it be better if we move all tracking constants to a single class file? This way we can avoid duplications or name changes between screens.

Sounds good @Balcan , do you think is it better to have a analytics constant class per presenter instead ?

Balcan:
> > My only comment. Wouldn't it be better if we move all tracking constants to a single class file? This way we can avoid duplications or name changes between screens.
> 
> Sounds good @Balcan , do you think is it better to have a analytics constant class per presenter instead ? Or just one?

O would have just one class. We can discuss this later in the daily. I was thinking on having different access layers

> > My only comment. Wouldn't it be better if we move all tracking constants to a single class file? This way we can avoid duplications or name changes between screens.
> 
> Sounds good @Balcan , do you think is it better to have a analytics constant class per presenter instead ? Or just one?

We can discuss it later in the daily. But I was thinking on having some kind of provider so that we can group the constants by screen, category and action and use them like:
`Analytics.Screen.Home, Analytics.Category.SyncButton, Analytics.Action.Click`

JaimeToca:
> > > My only comment. Wouldn't it be better if we move all tracking constants to a single class file? This way we can avoid duplications or name changes between screens.
> > 
> > 
> > Sounds good @Balcan , do you think is it better to have a analytics constant class per presenter instead ? Or just one?
> 
> O would have just one class. We can discuss this later in the daily. I was thinking on having different access layers
> 
> > > My only comment. Wouldn't it be better if we move all tracking constants to a single class file? This way we can avoid duplications or name changes between screens.
> > 
> > 
> > Sounds good @Balcan , do you think is it better to have a analytics constant class per presenter instead ? Or just one?
> 
> We can discuss it later in the daily. But I was thinking on having some kind of provider so that we can group the constants by screen, category and action and use them like:
> `Analytics.Screen.Home, Analytics.Category.SyncButton, Analytics.Action.Click`

Sounds good :+1: lets talk about it in the daily

PR #1970, opened 2021-03-15
feat: [ANDROAPP-3310] Handle openId logout

Balcan:
## Description

[ jira issue ](https://jira.dhis2.org/browse/ANDROAPP-3310)

## Solution description
## Covered unit test cases
- [ ] Smartphone Emulator
- [ ] Tablet Emulator
- [ ] Smartphone
- [ ] Tablet
- [ ] 4.4
- [ ] 5.X - 6.X
- [ ] 7.X
- [ ] 8.X
- [ ] 9.X - 10.X
- [ ] Other
## Checklist

### DIFF
--- a/app/src/main/java/org/dhis2/usescases/general/ActivityGlobalAbstract.java
+++ b/app/src/main/java/org/dhis2/usescases/general/ActivityGlobalAbstract.java
@@ -21,10 +21,11 @@ import androidx.core.content.ContextCompat;
 import com.google.android.material.dialog.MaterialAlertDialogBuilder;
 import com.google.firebase.analytics.FirebaseAnalytics;
 
+import org.dhis2.App;
 import org.dhis2.Bindings.ExtensionsKt;
 import org.dhis2.BuildConfig;
 import org.dhis2.R;
-import org.dhis2.data.server.OpenIdSession;
+import org.dhis2.data.server.ServerComponent;
 import org.dhis2.usescases.login.LoginActivity;
 import org.dhis2.usescases.main.MainActivity;
 import org.dhis2.usescases.splash.SplashActivity;
@@ -70,8 +71,6 @@ public abstract class ActivityGlobalAbstract extends AppCompatActivity
     public AnalyticsHelper analyticsHelper;
     @Inject
     public CrashReportController crashReportController;
-    @Inject
-    public OpenIdSession openIdSession;
     private PinDialog pinDialog;
     private boolean comesFromImageSource = false;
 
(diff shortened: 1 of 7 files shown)


mmmateos (approved):

JaimeToca:
Hello @Balcan the PR is fine. The only thing that concerns me is linking lifecycle to data layer which is something that belongs to activities/fragments. thoughts @mmmateos @andresmr @ferdyrod ?

app/src/main/java/org/dhis2/usescases/general/ActivityGlobalAbstract.java
```diff
@@ -88,6 +92,14 @@ public void setScreenName(String name) {
 
     @Override
     protected void onCreate(@Nullable Bundle savedInstanceState) {
+        ServerComponent serverComponent = ((App) getApplicationContext()).getServerComponent();
+        if (serverComponent != null) {
+            serverComponent.openIdSession().setSessionCallback(this, () -> {
```
andresmr, line 97: If we check this in the onCreate() we only check if the session is expired when the user navigates through activites, right?
Maybe could we check it in the on resume?.

app/src/main/java/org/dhis2/usescases/general/ActivityGlobalAbstract.java
```diff
@@ -88,6 +92,14 @@ public void setScreenName(String name) {
 
     @Override
     protected void onCreate(@Nullable Bundle savedInstanceState) {
+        ServerComponent serverComponent = ((App) getApplicationContext()).getServerComponent();
```
andresmr, line 95: instead of add the code to the base activity, as this is related with session, would it make sense to have some kind os sessionController or sessionService and inject it where we need it in order to have this logic isolated?

JaimeToca (approved):

app/src/main/java/org/dhis2/usescases/general/ActivityGlobalAbstract.java
Balcan: It is an observable, it just subscribes to it when the activity is created. So it will be listening the whole time.

app/src/main/java/org/dhis2/usescases/general/ActivityGlobalAbstract.java
Balcan: At first I implemented it with an Inject. But there were problems in the splash and login activity as they also extend the ActivityGlobalAbstract but the server component has not been created yet.

app/src/main/java/org/dhis2/usescases/general/ActivityGlobalAbstract.java
andresmr: you are right 🤦🏽

PR #2116, opened 2021-05-10
fix: [ANDROAPP-3903] check uid is not null when saving image

mmmateos:
## Description

## Solution description
## Covered unit test cases
- [ ] Smartphone Emulator
- [ ] Tablet Emulator
- [ ] Smartphone
- [ ] Tablet
- [ ] 4.4
- [ ] 5.X - 6.X
- [ ] 7.X
- [ ] 8.X
- [ ] 9.X - 10.X
- [ ] Other
## Checklist

### DIFF
--- a/app/src/main/java/org/dhis2/usescases/eventsWithoutRegistration/eventCapture/EventCapturePresenterImpl.java
+++ b/app/src/main/java/org/dhis2/usescases/eventsWithoutRegistration/eventCapture/EventCapturePresenterImpl.java
@@ -602,9 +602,7 @@ public class EventCapturePresenterImpl implements EventCaptureContract.Presenter
     @SuppressLint("CheckResult")
     @Override
     public void saveImage(String uuid, String filePath) {
-        if (uuid != null) {
-            valueStore.save(uuid, filePath).blockingFirst();
-        }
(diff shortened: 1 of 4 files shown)


app/src/main/java/org/dhis2/usescases/eventsWithoutRegistration/eventCapture/EventCapturePresenterImpl.java
```diff
@@ -602,7 +602,9 @@ public void displayMessage(String message) {
     @SuppressLint("CheckResult")
     @Override
     public void saveImage(String uuid, String filePath) {
-        valueStore.save(uuid, filePath).blockingFirst();
+        if (uuid != null) {
```
andresmr, line 605: What would happen if uid is null, image wouldn't be saved? is there any feedback to the user?

mmmateos: This only prevents the crash, the uid should never be null, but I was not able to reproduce it. If uid is null nothing would happen.

andresmr (approved):

andresmr:
Good job @mmmateos !

Balcan (approved):

PR #2176, opened 2021-06-21
feat: [ANDROAPP-4050] Render Radar chart

Balcan:
## Description
[ jira issue ](https://jira.dhis2.org/browse/ANDROAPP-4050)

## Solution description
## Covered unit test cases
- [ ] Smartphone Emulator
- [ ] Tablet Emulator
- [ ] Smartphone
- [ ] Tablet
- [ ] 4.4
- [ ] 5.X - 6.X
- [ ] 7.X
- [ ] 8.X
- [ ] 9.X - 10.X
- [ ] Other
## Checklist


ferdyrod (approved):

app/build.gradle
```diff
@@ -222,6 +222,8 @@ dependencies {
     implementation "androidx.legacy:legacy-support-v4:${libraries.legacysupport}"
     implementation "androidx.multidex:multidex:${libraries.multidex}"
     implementation "androidx.constraintlayout:constraintlayout:${libraries.constraintlayout}"
+    implementation "androidx.lifecycle:lifecycle-livedata-ktx:${libraries.lifecycle}"
+    implementation "androidx.lifecycle:lifecycle-viewmodel-ktx:${libraries.lifecycle}"
```
andresmr, line 226: I think that you don't need this in app, because we use it in commons

app/src/main/res/values/strings.xml
```diff
@@ -846,4 +846,6 @@
     <string name="conf_error_value_to_assign_option_set">Program rule %s is trying to assign value %s which does not belongs to option set %s</string>
     <string name="conf_error_value_option_set">Value for field %s does not belong to the option set in %s</string>
     <string name="action_do_not_show_again">Do not show again</string>
+    <string name="feature_configuration">Feature Configuration</string>
```
andresmr, line 849: this string is declared in commons, please remove it

dhis_android_analytics/src/main/java/dhis2/org/analytics/charts/DhisAnalyticCharts.kt
```diff
@@ -11,10 +11,22 @@ class DhisAnalyticCharts @Inject constructor(
...
-    override fun getCharts(enrollmentUid: String): List<Graph> {
+    override fun geEnrollmentCharts(enrollmentUid: String): List<Graph> {
         return chartsRepository.getAnalyticsForEnrollment(enrollmentUid)
     }
 
+    override fun getProgramVisualizations(programUid: String): List<Graph> {
+        return chartsRepository.getAnalyticsForProgram(programUid)
+    }
+
+    override fun getHomeVisualizations(): List<Graph> {
```
andresmr, line 22: if we don't need this right now, why don't we declare them when needed?

andresmr (commented):
Why are you updating the SDK?

app/src/main/res/values/strings.xml
Balcan: I think transifex translations only fetch data from the main string.xml, so we need to overwrite (and duplicate) the keys in any module so that it can be translated. I have to make sure, but we may need to update how the translations are handled.

dhis_android_analytics/src/main/java/dhis2/org/analytics/charts/DhisAnalyticCharts.kt
Balcan: The idea was to have these charts (Pie and Radar) ready to be used by the visualizations in the search, home and dataSet screens... but we run into a blocker in the SDK (missing objects). They can be removed yes.

app/src/main/res/values/strings.xml
andresmr: Good catch 🧤 !

andresmr (approved):

PR #2178, opened 2021-06-22
feat: [ANDROAPP-4049] add pie chart for analytics

mmmateos:
## Description

## Solution description
## Covered unit test cases
- [ ] Smartphone Emulator
- [ ] Tablet Emulator
- [ ] Smartphone
- [ ] Tablet
- [ ] 4.4
- [ ] 5.X - 6.X
- [ ] 7.X
- [ ] 8.X
- [ ] 9.X - 10.X
- [ ] Other
## Checklist


commons/src/main/java/org/dhis2/commons/featureconfig/model/Feature.kt
```diff
@@ -1,5 +1,6 @@
 package org.dhis2.commons.featureconfig.model
 
 enum class Feature(val description: String) {
-    ANDROAPP_2557("Local Analytics 2.5")
+    ANDROAPP_2557("Local Analytics 2.5"),
+    ANDROAPP_4049("Pie Chart Analytics")
```
andresmr, line 5: I think that the idea using ANDROAPP_2557 which is the Epic of the Local Analytics is to gather together all the features of analytics, enabling and disabling it all in a row, so I would use ANDROAPP_2557 as well in you case.

dhis_android_analytics/src/main/java/dhis2/org/analytics/charts/DhisAnalyticCharts.kt
```diff
@@ -11,10 +11,22 @@ class DhisAnalyticCharts @Inject constructor(
...
         return chartsRepository.getAnalyticsForEnrollment(enrollmentUid)
     }
 
+    override fun getProgramVisualizations(programUid: String): List<Graph> {
+        return chartsRepository.getAnalyticsForProgram(programUid)
+    }
+
+    override fun getHomeVisualizations(): List<Graph> {
+        TODO("Not yet implemented")
+    }
```
andresmr, line 24: If we don't need these methods now, why do we declare them?

andresmr (commented):
are you updating the SDK in this commit?

Balcan (approved):

dhis_android_analytics/src/main/java/dhis2/org/analytics/charts/DhisAnalyticCharts.kt
Balcan: Copying the same answer from the radar chart PR:

The idea was to have these charts (Pie and Radar) ready to be used by the visualizations in the search, home and dataSet screens... but we run into a blocker in the SDK (missing objects).
I'm using them now in my next ticket... so If that's ok, i would leave them there.

commons/src/main/java/org/dhis2/commons/featureconfig/model/Feature.kt
mmmateos: Ah ok, I thought we have to add one feature for each ticket, I'll change it now

commons/src/main/java/org/dhis2/commons/featureconfig/model/Feature.kt
```diff
@@ -2,5 +2,4 @@ package org.dhis2.commons.featureconfig.model
 
 enum class Feature(val description: String) {
     ANDROAPP_2557("Local Analytics 2.5"),
```
andresmr, line 4: is there a missing comma?

andresmr (approved):

PR #2253, opened 2021-09-03
refactor: [ANDROAPP-4252] move filters to commons

mmmateos:
## Description

## Solution description
## Covered unit test cases
- [ ] Smartphone Emulator
- [ ] Tablet Emulator
- [ ] Smartphone
- [ ] Tablet
- [ ] 4.4
- [ ] 5.X - 6.X
- [ ] 7.X
- [ ] 8.X
- [ ] 9.X - 10.X
- [ ] Other
## Checklist


commons/build.gradle
```diff
@@ -55,6 +55,11 @@ android {
 dependencies {
     implementation fileTree(dir: "libs", include: ["*.jar"])
 
+    debugImplementation project(':core')
+    releaseImplementation("org.hisp.dhis:android-core:${libraries.dhis2sdk}") {
+        exclude group: "org.hisp.dhis", module: "core-rules"
+    }
+
```
ferdyrod, line 62: If we are adding the sdk dependency here, maybe it would a good idea to make available to the app module via the "releaseApi" and "debugApi" rather than having it duplicated in the App module. Maybe we can discuss it with the team.

commons/src/main/java/org/dhis2/commons/date/DateUtils.java
```diff
@@ -0,0 +1,232 @@
...
+
+import androidx.annotation.NonNull;
+import java.text.ParseException;
+import java.text.SimpleDateFormat;
+import java.util.Calendar;
+import java.util.Date;
+import java.util.Locale;
+
+public class DateUtils {
+
```
ferdyrod, line 11: Should we migrate this class to kotlin now that it in the commons module? Not in this issue but in a new one? What do you think?

commons/src/main/java/org/dhis2/commons/filters/data/FilterRepository.kt
```diff
@@ -1,24 +1,24 @@
-package org.dhis2.data.filter
+package org.dhis2.commons.filters.data
```
ferdyrod, line 2: When merging to develop there is another issue [ANDROAPP-3129](#2250) that adds to this class, We need to take that into consideration.

commons/build.gradle
Balcan: Yes, this could be a good idea :)

Balcan (approved):

commons/src/main/java/org/dhis2/commons/date/DateUtils.java
mmmateos: Yes I agree, with the DateUtils I think we should create a new issue to move all the class methods, and It would be great to migrate it to kotlin

PR #2416, opened 2021-11-16
refactor: [ANDROAPP-4431] remove picture view and viewmodel

mmmateos:
## Description

## Solution description
## Covered unit test cases
- [ ] Smartphone Emulator
- [ ] Tablet Emulator
- [ ] Smartphone
- [ ] Tablet
- [ ] 4.4
- [ ] 5.X - 6.X
- [ ] 7.X
- [ ] 8.X
- [ ] 9.X - 10.X
- [ ] Other
## Checklist


form/src/main/res/layout/form_picture.xml
```diff
@@ -6,22 +6,18 @@
...
-
-        <variable
-            name="description"
-            type="java.lang.String" />
+            name="item"
+            type="org.dhis2.form.model.FieldUiModel" />
 
     </data>
 
     <androidx.constraintlayout.widget.ConstraintLayout
```
andresmr, line 17: could we move this layout to form module?

form/src/main/java/org/dhis2/form/ui/FormViewModel.kt
```diff
@@ -173,7 +173,10 @@ class FormViewModel(
     ): Throwable? {
         return fieldValue?.let { value ->
             var error =
-                when (val result = valueType?.validator?.validate(value)) {
+                when (
+                    val result = valueType?.takeIf { it != ValueType.IMAGE }
```
andresmr, line 177: ValueType.IMAGE has the following validator:
` IMAGE(String::class.java, UidValidator);`
Is this class: UidValidator.kt. Why can not be validated?

form/src/main/res/layout/form_picture.xml
mmmateos: I think is already in the form module :smile:

form/src/main/java/org/dhis2/form/ui/FormViewModel.kt
mmmateos: The problem is when reaching this validator the value of image field is not a UID, is the path of the file that has to be stored. Maybe we could discuss if this validator could be skipped or we should make something else.

PR #2548, opened 2022-03-07
feat: [ANDROAPP-4591] data entry bottom dialog new design

mmmateos:
## Description

## Solution description
## Covered unit test cases
- [ ] Smartphone Emulator
- [ ] Tablet Emulator
- [ ] Smartphone
- [ ] Tablet
- [ ] 4.4
- [ ] 5.X - 6.X
- [ ] 7.X
- [ ] 8.X
- [ ] 9.X - 10.X
- [ ] Other
## Checklist

### DIFF
--- a/app/src/main/java/org/dhis2/ui/DataEntryBottomDialogContent.kt
+++ b/app/src/main/java/org/dhis2/ui/DataEntryBottomDialogContent.kt
@@ -26,8 +26,10 @@ import androidx.compose.ui.graphics.Color
 import androidx.compose.ui.res.colorResource
 import androidx.compose.ui.res.painterResource
 import androidx.compose.ui.res.stringResource
+import androidx.compose.ui.text.style.TextAlign
 import androidx.compose.ui.tooling.preview.Preview
 import androidx.compose.ui.unit.dp
+import androidx.compose.ui.unit.sp
 import org.dhis2.R
 import org.dhis2.commons.data.FieldWithIssue
 import org.dhis2.commons.data.IssueType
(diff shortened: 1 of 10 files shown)


ferdyrod (approved):

app/src/main/java/org/dhis2/ui/DataEntryBottomDialogContent.kt
```diff
@@ -0,0 +1,207 @@
...
+import androidx.compose.ui.unit.dp
+import org.dhis2.R
+import org.dhis2.commons.data.FieldWithIssue
+import org.dhis2.commons.data.IssueType
+
+@Composable
+fun DataEntryBottomDialogContent(
+    isSaved: Boolean,
+    message: String,
+    fieldsWithIssues: List<FieldWithIssue>? = emptyList(),
```
andresmr, line 38: Just an obervation:
In my opinion these three properties could be a single immutable Ui model for example:
`data class DialogContentUiModel (
val message: String,
val fieldsWithIssues: List<FieldWithIssue>? = emptyList()
)`

Regarding the boolean is saved, i think it is adding presentation logic to the Composable class to selec one string or another and maybe is more complex for testing. I think that what I would do when creating the ui model, I would compute the logic and add two string as properties. So my proposed model would be:
`data class DialogContentUiModel (
val message: String,
val fieldsWithIssues: List<FieldWithIssue>? = emptyList(),
val icon: Int,
val message2: String
)`
I put message2 becuase I don't know the difference between message and message2, maybe it is title and subtitle.

I this way all logic can be model to a ViewModel or presentation component and be tested in once. Also with a single UI model we can, in the future assert our final expected result without espresso.

Please, let me know what do you think.

andresmr (approved):

JaimeToca (approved):

andresmr (approved):

andresmr (approved):

PR #2648, opened 2022-04-28
fix: [ANDROAPP-4718] save coordinates

mmmateos:
## Description

## Solution description
## Covered unit test cases
- [ ] Smartphone Emulator
- [ ] Tablet Emulator
- [ ] Smartphone
- [ ] Tablet
- [ ] 4.4
- [ ] 5.X - 6.X
- [ ] 7.X
- [ ] 8.X
- [ ] 9.X - 10.X
- [ ] Other
## Checklist

### DIFF
--- a/form/src/main/java/org/dhis2/form/ui/FormViewModel.kt
+++ b/form/src/main/java/org/dhis2/form/ui/FormViewModel.kt
@@ -24,7 +24,9 @@ import org.dhis2.form.model.FieldUiModel
 import org.dhis2.form.model.InfoUiModel
 import org.dhis2.form.model.RowAction
 import org.dhis2.form.model.StoreResult
+import org.dhis2.form.model.UiRenderType
 import org.dhis2.form.model.ValueStoreResult
+import org.dhis2.form.ui.binding.getFeatureType
 import org.dhis2.form.ui.intent.FormIntent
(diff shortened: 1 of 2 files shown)


ferdyrod (approved):

form/src/main/java/org/dhis2/form/ui/FormViewModel.kt
```diff
@@ -190,42 +192,53 @@ class FormViewModel(
...
+    } ?: StoreResult(
+        rowAction.id,
+        ValueStoreResult.VALUE_HAS_NOT_CHANGED
+    )
 
-    fun valueTypeIsTextField(valueType: ValueType): Boolean {
+    fun valueTypeIsTextField(valueType: ValueType, renderType: UiRenderType? = null): Boolean {
         return valueType.isNumeric ||
-            valueType.isText ||
+            valueType.isText && renderType?.isPolygon() != true ||
```
andresmr, line 221: shouldn't has also `valueType.isText ||` ?

andresmr (approved):

form/src/main/java/org/dhis2/form/ui/FormViewModel.kt
mmmateos: No, for this case we need to return false if the coordinate is a polygon. If we add valueType.isText it will return true for the polygon case

PR #2666, opened 2022-05-10
fix: [ANDROAPP-4736] FormView: Could not find fragment crash

Balcan:
## Description

[ jira issue ](https://jira.dhis2.org/browse/ANDROAPP-4736)

## Solution description
## Covered unit test cases
- [ ] Smartphone Emulator
- [ ] Tablet Emulator
- [ ] Smartphone
- [ ] Tablet
- [ ] 4.4
- [ ] 5.X - 6.X
- [ ] 7.X
- [ ] 8.X
- [ ] 9.X - 10.X
- [ ] Other
## Checklist

### DIFF
--- a/app/src/main/java/org/dhis2/data/forms/dataentry/FormView.kt
+++ b/app/src/main/java/org/dhis2/data/forms/dataentry/FormView.kt
@@ -86,21 +86,21 @@ import org.hisp.dhis.android.core.common.ValueType
 import org.hisp.dhis.android.core.common.ValueTypeRenderingType
 import timber.log.Timber
 
-class FormView(
-    formRepository: FormRepository,
-    private val onItemChangeListener: ((action: RowAction) -> Unit)?,
-    private val locationProvider: LocationProvider?,
-    private val onLoadingListener: ((loading: Boolean) -> Unit)?,
-    private val onFocused: (() -> Unit)?,
-    private val onFinishDataEntry: (() -> Unit)?,
-    private val onActivityForResult: (() -> Unit)?,
-    private val needToForceUpdate: Boolean = false,
-    private val completionListener: ((percentage: Float) -> Unit)?,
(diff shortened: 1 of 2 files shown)


ferdyrod (approved):

app/src/main/java/org/dhis2/data/forms/dataentry/FormView.kt
```diff
@@ -187,6 +169,26 @@ class FormView(
...
+                    (parentFragment as FormViewConfigurationProvider).configuration()
+            }
+            else -> {
+                throw RuntimeException(
+                    "Activity or parent fragment must implement the " +
+                        "FormViewConfigurationProvider"
+                )
+            }
+        }
+    }
```
JaimeToca, line 191: I think the general idea of the configurator is good. Still, why attached to android activity?? we are imposing a restriction to everyone who wants to use the form to have a configurator in the activity that does not seem to be linked anywhere since you catch it with (context as // parent ass).configurator.

We have many options for example A think a simpler one might be returning a builder for the fragment and setting every listener in a similar way as before, you would not have a configurator but since it does not have any "logic" i am not sure about its purpose, the same could be done for the remaining fields. Thoughts @Balcan  @andresmr @ferdyrod  @mmmateos ???

app/src/main/java/org/dhis2/data/forms/dataentry/FormView.kt
```diff
@@ -823,6 +823,36 @@ class FormView(
         viewModel.loadData()
     }
 
+    fun setConfiguration(
```
andresmr, line 826: `setConfiguration` can be internal as it is used in the same packaged and does not need to be exposed.

andresmr (changes requested):

andresmr (approved):

PR #2676, opened 2022-05-13
fix: [ANDROAPP-4460] Transformation error from pie 

JaimeToca:
## Description
The problems comes from the transformation from pie chart to table. The SDK/Web is returning empty rows, therefore through the mapping periodId becomes empty string and therefore there is no position and end up being "-1". Other option should be to fix this in the SDK. Thoughts?


Balcan (approved):

ferdyrod (approved):

ferdyrod:
I agree that the SDK should address this.

PR #2726, opened 2022-06-01
feat: [Androapp-4766] user next logins not performs a sync

JaimeToca:
## Description

## Solution description
## Covered unit test cases
- [ ] Smartphone Emulator
- [ ] Tablet Emulator
- [ ] Smartphone
- [ ] Tablet
- [ ] 4.4
- [ ] 5.X - 6.X
- [ ] 7.X
- [ ] 8.X
- [ ] 9.X - 10.X
- [ ] Other
## Checklist


app/src/main/java/org/dhis2/usescases/main/MainPresenter.kt
```diff
@@ -257,4 +260,13 @@ class MainPresenter(
...
         )
     }
+
+    fun wasSyncAlreadyDone(): Boolean {
+        return syncIsPerformedInteractor.execute()
+    }
+
+    fun onDataSuccess() {
+        userManager.d2.dataStoreModule().localDataStore().value(WAS_INITIAL_SYNC_DONE)
+            .blockingSet("True")
```
Balcan, line 270: Should we move the hardcoded string to a const value in a LocalDataStoreKey class or something like that. This way we could avoid "true" != "True" in case we don't know exactly the value.

ferdyrod (approved):

app/src/main/java/org/dhis2/usescases/main/MainPresenter.kt
JaimeToca: Yes let me change it
Thanks @Balcan

PR #3243, opened 2023-07-26
fix: [ANDROAPP-5377 ] A TEI enrolled in many programs display to many icons blocking the TEI info card

xavimolloy:
## Description

[ jira issue ](https://dhis2.atlassian.net/browse/ANDROAPP-5377

## Solution description
## Covered unit test cases
- [ ] Smartphone Emulator
- [ ] Tablet Emulator
- [ ] Smartphone
- [ ] Tablet
- [ ] 4.4
- [ ] 5.X - 6.X
- [ ] 7.X
- [ ] 8.X
- [ ] 9.X - 10.X
- [ ] 11.X - 13.X
- [ ] Other
## Checklist

### DIFF
--- a/commons/src/main/java/org/dhis2/commons/bindings/TEICardExtensions.kt
+++ b/commons/src/main/java/org/dhis2/commons/bindings/TEICardExtensions.kt
@@ -28,6 +28,7 @@ import java.io.File
 import java.util.ArrayList
 import java.util.Date
 import org.dhis2.commons.R
+import org.dhis2.commons.data.EnrollmentIconData
 import org.dhis2.commons.data.SearchTeiModel
 import org.dhis2.commons.databinding.ItemFieldValueBinding
 import org.dhis2.commons.date.toUiText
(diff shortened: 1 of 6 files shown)


commons/src/main/java/org/dhis2/commons/bindings/TEICardExtensions.kt
```diff
@@ -54,42 +55,57 @@ fun List<Program>.addEnrollmentIcons(
...
+                            }
+                    }
                 }
             }
         }
     }
 }
 
+private fun getOtherEnrollmentsForTei(teiEnrollmentCount: Int): Int {
+    return if (teiEnrollmentCount - 4 > 99) 99 else teiEnrollmentCount - 4
```
Balcan, line 106: Extract the number to named constants

commons/src/main/java/org/dhis2/commons/bindings/TEICardExtensions.kt
```diff
@@ -54,42 +58,61 @@ fun List<Program>.addEnrollmentIcons(
     parent: ComposeView,
     currentProgram: String?
 ) {
+    val listSize = this.size
+    var programCount = 0
     parent.apply {
         setContent {
```
andresmr, line 64: I think that in this extension function there are mixing concerns as the presentation logic of calculating the icons to draw and number of remaining programs with the UI code. 
In my opinion, to make the function testable (we can create a `TEICardExtensionsTest` to test the function), you can split the extension and the UI returning in the extension function a data class, for example:
`data class EnrollmentIconData(val color: Color, val imageResource: Int)`
And the return the items in the extension function like:
`fun List<Program>.getEnrollmentIconsData(context: Context, currentProgram: String?): List<EnrollmentIconData>`
Now you can write unit test for the logic.
And where you are using the extension function you will receive a list of items to render so you can write the Ui code to render.
What do you think?, let me know if you have any doubt.

xavimolloy: I agree it is not a good idea to mix the presentation logic with the UI code. I will try and do it as you say. Thanks for the feedback!

ferdyrod (approved):

PR #3288, opened 2023-09-11
Androapp 5151 form display program stage section description

Balcan:
## Description

## Solution description
## Covered unit test cases
- [ ] Smartphone Emulator
- [ ] Tablet Emulator
- [ ] Smartphone
- [ ] Tablet
- [ ] 4.4
- [ ] 5.X - 6.X
- [ ] 7.X
- [ ] 8.X
- [ ] 9.X - 10.X
- [ ] 11.X - 13.X
- [ ] Other
## Checklist

### DIFF
--- a/commons/build.gradle.kts
+++ b/commons/build.gradle.kts
@@ -55,6 +55,10 @@ android {
     composeOptions {
         kotlinCompilerExtensionVersion = libs.versions.kotlinCompilerExtensionVersion.get()
     }
+
+    configurations.all {
+        resolutionStrategy.cacheDynamicVersionsFor(0, TimeUnit.SECONDS)
+    }
 }
 
 dependencies {
@@ -117,5 +121,7 @@ dependencies {
     api(libs.test.espresso.idlingconcurrent)
     api(libs.analytics.sentry)
     implementation(libs.github.treeView)
-    api(libs.dhis2.mobile.designsystem)
+    api(libs.dhis2.mobile.designsystem) {
+        isChanging = true
+    }
 }
--- a/form/src/main/java/org/dhis2/form/model/FormSection.kt
+++ b/form/src/main/java/org/dhis2/form/model/FormSection.kt
@@ -1,0 +1,16 @@
+package org.dhis2.form.model
+
+import org.hisp.dhis.mobile.ui.designsystem.component.SectionState
+
+data class FormSection(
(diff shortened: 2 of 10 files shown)


commons/build.gradle.kts
```diff
@@ -55,6 +55,10 @@ android {
     composeOptions {
         kotlinCompilerExtensionVersion = libs.versions.kotlinCompilerExtensionVersion.get()
     }
+
+    configurations.all {
+        resolutionStrategy.cacheDynamicVersionsFor(0, TimeUnit.SECONDS)
```
andresmr, line 60: does this refresh all the dependencies? or onlye the ones with `isChanging = true`?

form/src/main/java/org/dhis2/form/model/FormSection.kt
andresmr, line 1: there is this FormSection and a Section data class inside form, maybe the they are duplicated. I will keep it separate. What do you think?

form/src/main/java/org/dhis2/form/ui/Form.kt
```diff
@@ -195,23 +233,89 @@ private fun FieldProvider(
     needToForceUpdate: Boolean,
     textWatcher: TextWatcher,
     coordinateTextWatcher: LatitudeLongitudeTextWatcher,
+    uiEventHandler: (RecyclerViewUiEvents) -> Unit,
```
andresmr, line 236: I think that this property is unused, could we lost this interaction?. How are we going to implement it?

commons/build.gradle.kts
Balcan: only those whith the isChanging yes

form/src/main/java/org/dhis2/form/model/FormSection.kt
Balcan: The section data class was just a mock to allow me finish the implementation, but this is the one i was assuming you were changing/creating in your ticket :)

form/src/main/java/org/dhis2/form/ui/Form.kt
Balcan: It is not lost, Text value type does not uses it. When we implement Date in the design sytem it will.

andresmr (approved):

mmmateos (approved):

PR #3319, opened 2023-10-06
Event dashboard landscape

yuzalsif:
## Description
This is an improvement of the event dashboard layout for tablet view. It is based on the prototype on the figma design in the link below.

[Flow 2(Figma Design)](https://www.figma.com/proto/aPn2s7a2u5uaWYKtvlf4wi/Event-dashboard-landscape?page-id=2%3A5&node-id=414%3A81516&viewport=1055%2C0%2C0.07&scaling=scale-down&starting-point-node-id=414%3A81516&show-proto-sidebar=1)

- [ ] Smartphone Emulator
- [ ] Tablet Emulator
- [ ] Smartphone
- [ ] Tablet
- [ ] 4.4
- [ ] 5.X - 6.X
- [ ] 7.X
- [ ] 8.X
- [x] 9.X - 10.X
- [x] 11.X - 13.X
- [ ] Other
## Checklist


andresmr (changes requested):

andresmr:
Lint check is failing. You can check it via: `.gradlew ktlintCheck`also if you run `.gradlew ktlintFormat` you can autocorrect many format errors.

andresmr (changes requested):

app/src/main/java/org/dhis2/usescases/enrollment/EnrollmentActivity.kt
```diff
@@ -193,7 +193,7 @@ class EnrollmentActivity : ActivityGlobalAbstract(), EnrollmentView {
 
     override fun openEvent(eventUid: String) {
         if (presenter.isEventScheduleOrSkipped(eventUid)) {
-            val scheduleEventIntent = ScheduledEventActivity.getIntent(this, eventUid)
+            val scheduleEventIntent = ScheduledEventActivity.getIntent(this, eventUid, "", "")
```
andresmr, line 196: why in enrollment activity is a empty teiUid and enrolmmentUid?

app/src/main/java/org/dhis2/usescases/enrollment/EnrollmentActivity.kt
```diff
@@ -255,6 +257,7 @@ class EnrollmentActivity : ActivityGlobalAbstract(), EnrollmentView {
 
     @Deprecated("Deprecated in Java")
     override fun onBackPressed() {
+        super.onBackPressed()
```
andresmr, line 260: why do we need calling the super?

yuzalsif: I was getting this warning that "Overriding method should call super.onBackPressed" so that is why I added the call super. But I have removed it anyway.

app/src/main/java/org/dhis2/usescases/enrollment/EnrollmentActivity.kt
yuzalsif: I added optional field teiUid and enrollmentUid which can have default value of empty strings, now I have moved the default value for those fields in the definition of the function itself and no need to pass empty value when calling the function so I fixed that in the next commit.

PR #3321, opened 2023-10-09
ANDROAPP-5588-Implement-InputQRCode

xavimolloy:
## Description

## Solution description
## Covered unit test cases
- [ ] Smartphone Emulator
- [ ] Tablet Emulator
- [ ] Smartphone
- [ ] Tablet
- [ ] 4.4
- [ ] 5.X - 6.X
- [ ] 7.X
- [ ] 8.X
- [ ] 9.X - 10.X
- [ ] 11.X - 13.X
- [ ] Other
## Checklist


commons/src/main/res/values/strings.xml
```diff
@@ -214,6 +214,9 @@
     <string name="add_image">Add Image</string>
     <string name="add_signature">Add signature</string>
     <string name="add_file">Add file</string>
+    <string name="download">Download</string>
+    <string name="scan">Scan</string>
+    <string name="qr_code">QR code</string>
```
andresmr, line 219: As this texts are related to the form why not add the to the string in the form module?

andresmr (changes requested):

commons/src/main/res/values/strings.xml
xavimolloy: good idea, I've just updated with the changes

form/src/main/java/org/dhis2/form/ui/dialog/QRDetailBottomDialog.kt
```diff
@@ -26,34 +42,40 @@ import org.dhis2.form.data.FormFileProvider
...
     companion object {
         const val TAG: String = "QR_DETAIL_DIALOG"
     }
 
-    @Inject
-    lateinit var colorUtils: ColorUtils
-
     private lateinit var binding: QrDetailDialogBinding
     private var qrContentUri: Uri? = null
+    private var useComposeDialog: Boolean = useCompose
```
ferdyrod, line 72: I think there is no need to use another variable, you can directly use `useCompose`

xavimolloy: I tried that, but it doesn't find the variable from the onCreateView method

andresmr: Maybe because you need `**private val** useCompose: Boolean` in line 60?

form/src/main/java/org/dhis2/form/ui/provider/inputfield/FieldProvider.kt
```diff
@@ -263,29 +265,41 @@ private fun ProvideInputText(
...
-            value = it
-            intentHandler(
-                FormIntent.OnTextChange(
-                    fieldUiModel.uid,
-                    value,
-                    fieldUiModel.valueType,
-                ),
+    when (fieldUiModel.renderingType) {
+        UiRenderType.QR_CODE -> {
+            ProvideQRInput(
```
andresmr, line 272: I would not call a provide from another provide, maybe you can check this in line 77, if ValueType is Text, then you check the render type and call provideText or provideQR, what do you think?

andresmr (changes requested):

form/src/main/java/org/dhis2/form/ui/provider/inputfield/FieldProvider.kt
xavimolloy: That's how I initially set it up but the method got flagged by Sonar for having too much complexity if I managed it there

form/src/main/java/org/dhis2/form/ui/dialog/QRDetailBottomDialog.kt
xavimolloy: that was it! thanks I've updated the PR

andresmr (approved):

PR #3327, opened 2023-10-10
Androapp 5485 tei event and dataset card redesign

andresmr:


app/src/main/java/org/dhis2/usescases/datasets/datasetDetail/datasetList/mapper/DatasetCardMapper.kt
```diff
@@ -0,0 +1,234 @@
...
+                    value = resourceManager.getString(R.string.not_synced),
+                    color = AdditionalInfoItemColor.DISABLED.color,
+                    isConstantItem = true,
+                )
+            }
+
+            State.UPLOADING -> {
+                AdditionalInfoItem(
+                    icon = {
+                        ProgressIndicator(type = ProgressIndicatorType.CIRCULAR)
```
xavimolloy, line 124: Verify this with Marcos but I believe he wanted us to use the same Icon as the sync button, because of the problem with the width of the progress indicator at that size

app/src/main/java/org/dhis2/usescases/programEventDetail/ProgramEventDetailLiveAdapter.kt
```diff
@@ -35,9 +42,41 @@ class ProgramEventDetailLiveAdapter(
...
 
     override fun onBindViewHolder(holder: EventViewHolder, position: Int) {
-        holder.bind(getItem(position), null) {
-            getItem(holder.bindingAdapterPosition)?.toggleValueList()
-            notifyItemChanged(holder.bindingAdapterPosition)
+        getItem(position)?.let {
+            val materialView = holder.itemView.findViewById<ConstraintLayout>(R.id.materialView)
+            materialView.visibility = View.GONE
+            val composeView = holder.itemView.findViewById<ComposeView>(R.id.composeView)
+            composeView.setContent {
```
xavimolloy, line 49: will we still allow the option of using the old design here?

xavimolloy (approved):
Great work!

app/src/main/java/org/dhis2/usescases/datasets/datasetDetail/datasetList/mapper/DatasetCardMapper.kt
andresmr: Yes, I saw it and we talked, I forgot to change it.  I'll change it, thanks!.

app/src/main/java/org/dhis2/usescases/programEventDetail/ProgramEventDetailLiveAdapter.kt
andresmr: No, we are not, neither in dataset. In search is different because we reuse it for relationships, which we'll need to change in the future. I didn't remove the adapters, holders and recyclerviews because the issue will be too big and too many changes. I replace just the item. My idea is to create separate issues to remove all xml vies in these screens.

app/src/main/java/org/dhis2/usescases/searchTrackEntity/ui/mapper/TEICardMapper.kt
```diff
@@ -0,0 +1,371 @@
...
+    private fun checkEnrolledPrograms(
+        list: MutableList<AdditionalInfoItem>,
+        enrolledPrograms: List<Program>,
+    ) {
+        val programNames = enrolledPrograms.map { it.name() }
+
+        if (programNames.size > 1) {
+            list.add(
+                AdditionalInfoItem(
+                    key = "Programs:",
```
mmmateos, line 217: Shouldn't be a string resource to be translated?

mmmateos (approved):

PR #3346, opened 2023-10-20
[ANDROAPP-5603] Input coordinate

andresmr:


xavimolloy (approved):

form/src/main/java/org/dhis2/form/ui/provider/inputfield/CoordinatesProvider.kt
```diff
@@ -0,0 +1,76 @@
...
+                    RecyclerViewUiEvents.RequestLocationByMap(
+                        uid = fieldUiModel.uid,
+                        featureType = FeatureType.POINT,
+                        value = fieldUiModel.value,
+                    ),
+                )
+            },
+        )
+    } else {
+        // TODO(ImplementPOlygon)
```
mmmateos, line 60: There are two checks of rendertype point or polygon, here and in FieldProvider, we should only have one

andresmr: solved

mmmateos (approved):

PR #3405, opened 2023-11-22
fix: [ANDROAPP-5741] App crashes if charts are empty

xavimolloy:
## Description

## Solution description
## Covered unit test cases
- [ ] Smartphone Emulator
- [ ] Tablet Emulator
- [ ] Smartphone
- [ ] Tablet
- [ ] 4.4
- [ ] 5.X - 6.X
- [ ] 7.X
- [ ] 8.X
- [ ] 9.X - 10.X
- [ ] 11.X - 13.X
- [ ] Other
## Checklist

### DIFF
--- a/dhis_android_analytics/src/main/java/dhis2/org/analytics/charts/data/Graph.kt
+++ b/dhis_android_analytics/src/main/java/dhis2/org/analytics/charts/data/Graph.kt
@@ -127,7 +127,6 @@ data class Graph(
             PeriodType.FinancialNov,
             -> {
                 val date = minDate().plusYears(numberOfSteps)
-
                 YearMonth.from(date).atDay(1)
             }
         }


dhis_android_analytics/src/main/java/dhis2/org/analytics/charts/data/Graph.kt
```diff
@@ -24,13 +24,32 @@ data class Graph(
...
+
+            return minDate
+        } else { LocalDate.now() }
+    }
 
-    private fun maxDate() = series.maxOfOrNull { serie ->
-        serie.coordinates.maxOf { point -> point.eventDate }
-    }?.toInstant()?.atZone(ZoneId.systemDefault())?.toLocalDate() ?: LocalDate.now()
+    private fun maxDate(): LocalDate? {
+        val coordinatesNotEmpty = series.maxOfOrNull { serie ->
```
mmmateos, line 41: Couldn't using maxOfOrNull to determine a boolean give us wrong data?

dhis_android_analytics/src/main/java/dhis2/org/analytics/charts/data/Graph.kt
```diff
@@ -116,7 +135,7 @@ data class Graph(
             PeriodType.SixMonthlyApril,
             PeriodType.SixMonthlyNov,
             -> {
-                val date = minDate().plusMonths(numberOfSteps)
+                val date = minDate()?.plusMonths(numberOfSteps)
```
andresmr, line 138: need to be minDate nullable? could this lead to a null pointer in the next line?

xavimolloy: No as the method returns the local date if anything is null, although I understand it could lead to confusion so I have updated the method to better reflect the return value, it is not nullable anymore

dhis_android_analytics/src/main/java/dhis2/org/analytics/charts/data/Graph.kt
xavimolloy: The problem is that the check is done over a list of SerieDate that contains a list of Coordinates. I couldn't simply add minOfOrNull to the nested check as it formats the value later. otherwise it will always return Local date so I believe correct output will be maintained with the new changes. it will return localDate if coordinate list is empty.

andresmr (approved):

ferdyrod (approved):

andresmr (approved):

mmmateos (approved):

PR #3459, opened 2024-01-10
Androapp 5724 remove organization unit in tei event ds card if user only has access to one org unit

siddh1004:
## Description
IF the user has access to more than one org unit in capture OR search tree  - DISPLAY ORG UNIT IN CARD

IF the user has access to only ONE org unit in capture AND search tree  - DON’T DISPLAY ORG UNIT (Enrolled in/Registered in)

[ jira issue ](https://dhis2.atlassian.net/browse/ANDROAPP-5724)

## Solution description
## Covered unit test cases
- [x] Smartphone Emulator
- [ ] Tablet Emulator
- [ ] Smartphone
- [ ] Tablet
- [ ] 4.4
- [ ] 5.X - 6.X
- [ ] 7.X
- [ ] 8.X
- [ ] 9.X - 10.X
- [x] 11.X - 13.X
- [ ] Other
## Checklist

### DIFF
--- a/app/src/main/java/org/dhis2/usescases/programEventDetail/ProgramEventMapper.kt
+++ b/app/src/main/java/org/dhis2/usescases/programEventDetail/ProgramEventMapper.kt
@@ -5,7 +5,6 @@ import org.dhis2.commons.data.EventViewModel
 import org.dhis2.commons.data.EventViewModelType
 import org.dhis2.commons.data.ProgramEventViewModel
 import org.dhis2.commons.data.tuples.Pair
-import org.dhis2.commons.reporting.CrashReportController
 import org.dhis2.data.dhislogic.DhisPeriodUtils
 import org.dhis2.utils.DateUtils
 import org.hisp.dhis.android.core.D2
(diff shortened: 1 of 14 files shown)


app/src/main/java/org/dhis2/usescases/programEventDetail/ProgramEventMapper.kt
```diff
@@ -32,6 +32,10 @@ class ProgramEventMapper @Inject constructor(
         val programStage =
             d2.programModule().programStages().uid(event.programStage()).blockingGet()
 
+        val dataSetOrgUnitNumber = d2.organisationUnitModule().organisationUnits()
+            .byProgramUids(listOf(event.program()!!))
+            .blockingGet().size
+
         crashReportController.addBreadCrumb(
```
andresmr, line 39: please, can you remove this breadcrumb, we no longer need it.

app/src/main/java/org/dhis2/usescases/programEventDetail/ProgramEventMapper.kt
```diff
@@ -62,6 +66,7 @@ class ProgramEventMapper @Inject constructor(
             },
             nameCategoryOptionCombo =
             getCategoryComboFromOptionCombo(event.attributeOptionCombo())?.displayName(),
+            displayOrgUnitName = dataSetOrgUnitNumber > 1,
```
andresmr, line 69: EventViewModel is a Domain model, indeed it is in `package org.dhis2.commons.data`, `displayOrgUnitName` is more related to presentation logic, so I will evaluate it in a UseCase as you are going to reuse it, and the inject the use case in the classes where you generate the UI model like `ListCardUiModel`. 
You can follow the [Android guidelines](https://developer.android.com/topic/architecture/domain-layer) to generate use cases.

andresmr (changes requested):

app/src/main/java/org/dhis2/usescases/programEventDetail/ProgramEventMapper.kt
siddh1004: breadcrumb removed

app/src/main/java/org/dhis2/usescases/programEventDetail/ProgramEventMapper.kt
siddh1004: `displayOrgUnitName` moved to UI model

andresmr (approved):

mmmateos (approved):

PR #3462, opened 2024-01-10
ANDROAPP-5829-adapt-input-image-to-new-mobile-ui-functionality

msasikanth:


app/src/main/java/org/dhis2/usescases/searchTrackEntity/listView/SearchTEList.kt
```diff
@@ -232,8 +231,13 @@ class SearchTEList : FragmentGlobalAbstract() {
 
     // TODO replace ImageDetailBottomDialog with new mobile ui Image detail component
```
andresmr, line 232: please remove this line

form/src/main/java/org/dhis2/form/ui/FormView.kt
```diff
@@ -963,8 +974,13 @@ class FormView : Fragment() {
 
     // TODO replace ImageDetailBottomDialog with new mobile ui Image detail component
```
andresmr, line 975: Please, remove this commented line

app/src/main/java/org/dhis2/usescases/searchTrackEntity/mapView/SearchTEMap.kt
```diff
@@ -260,14 +259,15 @@ class SearchTEMap : FragmentGlobalAbstract(), MapboxMap.OnMapClickListener {
...
                 if (binding.mapCarousel.carouselEnabled) {
-                    // TODO replace ImageDetailBottomDialog with new mobile ui Image detail component
-                    ImageDetailBottomDialog(
-                        null,
-                        File(path),
-                    ).show(
-                        childFragmentManager,
-                        ImageDetailBottomDialog.TAG,
-                    )
+                    if (!path.isNullOrBlank()) {
```
andresmr, line 262: this nested if is generating a code smell, maybe you can merge them

commons/src/main/java/org/dhis2/commons/dialogs/imagedetail/ImageDetailActivity.kt
```diff
@@ -0,0 +1,79 @@
...
+                title = title.orEmpty(),
+                onDismiss = { finish() },
+                onDownloadButtonClick = { saveImage(imagePath) },
+                onShareButtonClick = { shareImage(imagePath) },
+            )
+        }
+    }
+
+    private fun saveImage(image: String) {
+        // TODO: Implement saving image to disk
```
andresmr, line 56: I would develop a new download image implementation given that the one in Qr it seem is not working properly, and then this new implementation could be reused in other components like QR one. What do you think?

andresmr (changes requested):

commons/src/main/java/org/dhis2/commons/dialogs/imagedetail/ImageDetailActivity.kt
msasikanth: Yeah, we would have to save the image publicly accessible folder like downloads. That should make the implementation simpler. We can do this in another story?

We also need to replace it in other places where the `InputImage` component is used as well (in Forms?), so we can do replace all the save image functionality in a single story. WDYT?

andresmr: yes, it is ok. so we could merge this without implementation and the develop the other one separately

andresmr (approved):

xavimolloy (approved):

PR #3525, opened 2024-03-10
Update SearchTETest

andresmr:
## Description
Update SearchTETest class with lazynActivityScenarioRule and composeTestRule


Jenkinsfile
andresmr, line 1: I removed these steps due to they are performed in GithubActions

PR #3535, opened 2024-03-13
ANDROAPP-5799-search-enroll-button

msasikanth:
~~Unfortunately the query data is not observable, so i couldn't observe it to make changes to the UI to hide/show the new button. We need to make it observable to do it.~~


msasikanth:
> ~Unfortunately the query data is not observable, so i couldn't observe it to make changes to the UI to hide/show the new button. We need to make it observable to do it.~

While the query data is not mutable, the UI state is. So, I decided to use that as key for composables to recompose them and get the latest query data whenever UI state changes. Since it's not a disk or expensive operation, it shouldn't causing any performance issues.

xavimolloy (approved):

msasikanth:
@andresmr the failing test here passed locally, not sure if that is a flaky one.

andresmr (approved):

PR #3553, opened 2024-03-19
fi: [ANDROAPP-6004] All {TE Type} search now show too many result.

ferdyrod:
## Description

## Solution description
## Covered unit test cases
- [ ] Smartphone Emulator
- [ ] Tablet Emulator
- [ ] Smartphone
- [ ] Tablet
- [ ] 4.4
- [ ] 5.X - 6.X
- [ ] 7.X
- [ ] 8.X
- [ ] 9.X - 10.X
- [ ] 11.X - 13.X
- [ ] Other
## Checklist

### DIFF
--- a/app/src/main/java/org/dhis2/usescases/searchTrackEntity/ui/mapper/TEICardMapper.kt
+++ b/app/src/main/java/org/dhis2/usescases/searchTrackEntity/ui/mapper/TEICardMapper.kt
@@ -129,8 +129,13 @@ class TEICardMapper(
                 list = list,
                 enrolledPrograms = searchTEIModel.programInfo,
             )
+            val programUid: String = if (searchTEIModel.selectedEnrollment != null) {
+                searchTEIModel.selectedEnrollment.program().toString()
+            } else {
+                ""
+            }
             checkEnrollmentStatus(
-                programUid = searchTEIModel.selectedEnrollment.program()!!,
+                programUid = programUid,
                 list = list,
                 status = searchTEIModel.selectedEnrollment?.status(),
             )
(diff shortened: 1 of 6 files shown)


app/src/main/java/org/dhis2/usescases/searchTrackEntity/ui/mapper/TEICardMapper.kt
```diff
@@ -129,8 +129,13 @@ class TEICardMapper(
                 list = list,
                 enrolledPrograms = searchTEIModel.programInfo,
             )
+            val programUid: String = if (searchTEIModel.selectedEnrollment != null) {
+                searchTEIModel.selectedEnrollment.program().toString()
+            } else {
+                ""
```
andresmr, line 135: which are the consecuences of having programUid as an emtpy string?

mmmateos (approved):

app/src/main/java/org/dhis2/usescases/searchTrackEntity/ui/mapper/TEICardMapper.kt
ferdyrod: This is use to get a custom enrollmentLabel if there is one. If this is empty it will have the same impact as if it is null. It will show the default string "enrollment(s)"

andresmr: so, what do you think about setting it as null if we don't have a prograUid, I think considering empty string as null might cause unexpected flows that can be invisible for us.  What do you think?

ferdyrod: It's ok, I'll change so that the method checkEnrollmentStatus accepts null value as programUid

Balcan (approved):

PR #3558, opened 2024-03-20
ANDROAPP-5712-sync-stops-when-users-click-outside-the-sync-pop-up-card

msasikanth:


app/src/main/java/org/dhis2/utils/granularsync/SyncStatusDialog.kt
```diff
@@ -94,23 +99,37 @@ class SyncStatusDialog : BottomSheetDialogFragment(), GranularSyncContracts.View
         container: ViewGroup?,
         savedInstanceState: Bundle?,
     ): View {
-        return ComposeView(requireContext()).apply {
-            setViewCompositionStrategy(ViewCompositionStrategy.DisposeOnViewTreeLifecycleDestroyed)
-            setContent {
-                Mdc3Theme {
-                    val syncState by viewModel.currentState.collectAsState()
-                    syncState?.let { syncUiState ->
+        viewLifecycleOwner.lifecycleScope.launch {
```
andresmr, line 102: Can you explain why are we observing `currentState` here,  I see when user interacts with the dialog the panel is dismissed, so why collecting it here?

msasikanth: We are observing it to react to the sync state changes and show snackbar or enable/disable touch outside to dismiss the sheet.

We were previously doing this in the composable with `collectAsState`, but that `when` block will get triggered every time there is a recomposition, which is not ideal. So, moved it outside to a proper lifecycle observable instead.

andresmr (approved):

mmmateos (approved):

Balcan:
Closing the PR. The sync process is working even if the dialog is closed. The problem is that it is not updated properly if reopened.

PR #3571, opened 2024-04-03
fix: [ANDROAPP-6040] Remove scroll to first empty item functionality

xavimolloy:
## Description

## Solution description
## Covered unit test cases
- [x] Smartphone Emulator
- [ ] Tablet Emulator
- [ ] Smartphone
- [ ] Tablet
- [ ] 4.4
- [ ] 5.X - 6.X
- [ ] 7.X
- [ ] 8.X
- [ ] 9.X - 10.X
- [x] 11.X - 13.X
- [ ] Other
## Checklist

### DIFF
--- a/form/src/main/java/org/dhis2/form/ui/Form.kt
+++ b/form/src/main/java/org/dhis2/form/ui/Form.kt
@@ -22,7 +22,6 @@ import androidx.compose.material.icons.outlined.ErrorOutline
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.LaunchedEffect
 import androidx.compose.runtime.MutableState
-import androidx.compose.runtime.derivedStateOf
 import androidx.compose.runtime.mutableStateOf
 import androidx.compose.runtime.remember
 import androidx.compose.runtime.rememberCoroutineScope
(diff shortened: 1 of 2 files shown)


form/src/main/java/org/dhis2/form/ui/Form.kt
```diff
@@ -90,14 +89,6 @@ fun Form(
...
                 key = { _, fieldUiModel -> fieldUiModel.uid },
             ) { _, section ->
-                val isSectionOpen = remember(section.state) {
-                    derivedStateOf { section.state == SectionState.OPEN }
-                }
-
-                LaunchIfTrue(isSectionOpen.value) {
-                    scrollState.animateScrollToItem(sections.indexOf(section))
-                    focusManager.moveFocusNext(focusNext)
-                }
```
ferdyrod, line 100: I may be wrong and we can discussed this with the team, but I think this should optional and not be removed completely

xavimolloy: Do you mean configurable the ASWA? I think that should be addressed in a different issue if so.

andresmr: I think that this is for all cases. 
There is a use case where we need this for, but with the current form is broken. It's for error handling, when user taps on sync conflicted item it should navigate to the form and scroll to the selected item, as it is not working maybe we can remove this for now and think about how to do it when we fix the other issue: 
image

andresmr (approved):

form/src/main/java/org/dhis2/form/ui/Form.kt
xavimolloy: I did take this into account but as Andres mentions it's not working currently with new form, and the part that I have removed checks for first open section, rather than any error field. I will create an issue to address the sync conflict item.

ferdyrod: Maybe I'm thinking far ahead. Since we're planing to release the form library at some point, would be like to stop automatically scrolling to an open section on all forms?.

xavimolloy: Maybe we could have a conversation with design we want to allow this or not. In this issue I could maybe add it in as a variable similarly to how we did it with the useCompose one. @marcamsn  What do you think?

ferdyrod (approved):

PR #3651, opened 2024-05-22
test: [ANDROAPP-6609] Refactor program rules UI form's test to integration tests

ferdyrod:
## Description

## Solution description
## Covered unit test cases
- [ ] Smartphone Emulator
- [ ] Tablet Emulator
- [ ] Smartphone
- [ ] Tablet
- [ ] 4.4
- [ ] 5.X - 6.X
- [ ] 7.X
- [ ] 8.X
- [ ] 9.X - 10.X
- [ ] 11.X - 13.X
- [ ] Other
## Checklist


form/build.gradle.kts
```diff
@@ -53,6 +53,12 @@ android {
     composeOptions {
         kotlinCompilerExtensionVersion = libs.versions.kotlinCompilerExtensionVersion.get()
     }
+
+    testOptions {
+        unitTests {
+            isReturnDefaultValues = true
```
andresmr, line 59: what is this for?

ferdyrod: This was added because there was a error saying that the method "isEmpty()"in the Android's TextUtils class was not being mocked. Even though that method is not invoked anywhere in the code. The error happened when the method FormViewModel::processCalculatedItems()  called the FormCountingIdlingResource.increment() method.

Adding this configuration fixed this problem

andresmr (approved):

app/src/androidTest/java/org/dhis2/usescases/filters/FilterTest.kt
```diff
@@ -146,12 +145,13 @@ class FilterTest : BaseTest() {
         eventWithoutRegistrationRobot(composeTestRule) {
             clickOnEventAtPosition(0)
         }
-        formRobot(composeTestRule) {
+        // Commented because FormRobot class wah eliminated as FormTest was also eliminated
+        /*formRobot(composeTestRule) {
```
xavimolloy, line 149: if this is no longer needed maybe we can remove it?

xavimolloy (approved):

app/src/androidTest/java/org/dhis2/usescases/filters/FilterTest.kt
ferdyrod: I don't think so, because this test is going to be looked at at some point. So it's better to leave it as it is so when someone does look into this, knows how the tests really is coded.

PR #3688, opened 2024-06-14
fix: [ANDROAPP-6132] double tap on event/enrollment creation could generate duplicates

xavimolloy:
## Description

## Solution description
## Covered unit test cases
- [ ] Smartphone Emulator
- [x] Tablet Emulator
- [ ] Smartphone
- [ ] Tablet
- [ ] 4.4
- [ ] 5.X - 6.X
- [ ] 7.X
- [ ] 8.X
- [ ] 9.X - 10.X
- [x] 11.X - 13.X
- [ ] Other
## Checklist


app/src/main/java/org/dhis2/usescases/searchTrackEntity/SearchTEActivity.java
```diff
@@ -506,13 +508,16 @@ private void observeDownload() {
     }
 
     private void observeLegacyInteractions() {
+        SingleEventEnforcer singleEventEnforcer = new SingleEventEnforcerImpl();
+
         viewModel.getLegacyInteraction().observe(this, legacyInteraction -> {
             if (legacyInteraction != null) {
                 switch (legacyInteraction.getId()) {
-                    case ON_ENROLL_CLICK -> {
+                    case ON_ENROLL_CLICK -> singleEventEnforcer.processEvent( () -> {
```
andresmr, line 516: What do you think about move this action to the presenter -> onEnrollClick(...)?
What is the purpose of _LegacyInteraction.OnEnrollClick interaction = (LegacyInteraction.OnEnrollClick) legacyInteraction;_

app/src/main/java/org/dhis2/usescases/teiDashboard/dashboardfragments/teidata/teievents/StageViewHolder.kt
```diff
@@ -93,7 +95,7 @@ internal class StageViewHolder(
                             },
                     ) {
                         NewEventOptions(presenter.getNewEventOptionsByStages(stage)) {
-                            presenter.onAddNewEventOptionSelected(it, stage)
+                            singleEventEnforcer.processEvent { presenter.onAddNewEventOptionSelected(it, stage) }
```
andresmr, line 98: Same here, could this be handled in the presenter?

commons/src/main/java/org/dhis2/commons/orgunitselector/OUTreeFragment.kt
```diff
@@ -165,7 +169,9 @@ class OUTreeFragment private constructor() : DialogFragment() {
     private fun showAsDialog() = arguments?.getBoolean(ARG_SHOW_AS_DIALOG, false) ?: false
 
     private fun confirmOuSelection() {
-        selectionCallback(presenter.getOrgUnits())
+        singleEventEnforcer.processEvent {
+            selectionCallback(presenter.getOrgUnits())
```
andresmr, line 173: Same here, I think this singleevent enforcer should be handled from presentation logic

andresmr (changes requested):

app/src/main/java/org/dhis2/usescases/searchTrackEntity/SearchTEActivity.java
xavimolloy: if you look at the code it is casting the event to the type of action on the when. it is present in all the options of Legacy interaction, if you can think of a more legible way of putting this I'm happy to update it. About moving the action to the presenter I think it makes sense for some of them but not for all, I will make my case in each of the comments

app/src/main/java/org/dhis2/usescases/teiDashboard/dashboardfragments/teidata/teievents/StageViewHolder.kt
xavimolloy: I agree and have updated this

commons/src/main/java/org/dhis2/commons/orgunitselector/OUTreeFragment.kt
xavimolloy: Here I believe that this should not be migrated to the presenter, since all the orgtree presenter is doing is return the selected org units, and this way if we trigger the callback from here, it will apply for all places the orgUnitDialog is called, otherwise we will have to adapt it in all the implementations.  Let me know if this makes sense to you

andresmr: Ok, I have checked the code. I think this is the unique usage apart from the testing, right?.

andresmr (approved):

Balcan (approved):

PR #3730, opened 2024-07-16
[ANDROAPP-6326] set default target SDK version

andresmr:


app/src/androidTest/java/org/dhis2/usescases/flow/syncFlow/robot/SyncFlowRobot.kt
```diff
@@ -30,9 +30,11 @@ class SyncFlowRobot(val composeTestRule: ComposeTestRule) : BaseRobot() {
         composeTestRule.onNodeWithTag(MAIN_BUTTON_TAG, useUnmergedTree = true).performClick()
     }
 
+    @OptIn(ExperimentalTestApi::class)
     fun checkSyncWasSuccessfully() {
         val expectedTitle = InstrumentationRegistry.getInstrumentation()
             .targetContext.getString(R.string.sync_dialog_title_synced)
+        composeTestRule.waitUntilAtLeastOneExists(hasText(expectedTitle))
```
xavimolloy, line 37: this method I believe has a default time that throws an exception if not found, and is 1000 ms if not specified. I have seen tests fail due to the ocasional slow loading of the page other times this assertion was used, do you think it's a good idea to increase the time out to 2000 ms at least?

mmmateos (approved):

app/src/androidTest/java/org/dhis2/usescases/flow/syncFlow/robot/SyncFlowRobot.kt
andresmr: yes, it has a default time 1000ms, so it's the time that the workers are setting sync successful, why don't we try the default implementation and if it fails we can think on increment it?

xavimolloy (approved):

PR #3737, opened 2024-07-22
feat: [ANDROAPP-5947] custom label for event at program level

mmmateos:
## Description

## Solution description
## Covered unit test cases
- [ ] Smartphone Emulator
- [ ] Tablet Emulator
- [ ] Smartphone
- [ ] Tablet
- [ ] 4.4
- [ ] 5.X - 6.X
- [ ] 7.X
- [ ] 8.X
- [ ] 9.X - 10.X
- [ ] 11.X - 13.X
- [ ] Other
## Checklist

### DIFF
--- a/app/src/main/java/org/dhis2/data/dhislogic/DhisProgramUtils.kt
+++ b/app/src/main/java/org/dhis2/data/dhislogic/DhisProgramUtils.kt
@@ -161,7 +161,7 @@ class DhisProgramUtils @Inject constructor(val d2: D2) {
             ProgramType.WITH_REGISTRATION -> {
                 program.trackedEntityType()?.displayName() ?: defaultTrackerLabel
             }
-            ProgramType.WITHOUT_REGISTRATION -> defaultEventLabel
+            ProgramType.WITHOUT_REGISTRATION -> program.eventLabel() ?: defaultEventLabel
             null -> ""
         }
     }
--- a/app/src/main/java/org/dhis2/data/dhislogic/DhisProgramUtils.kt
+++ b/app/src/main/java/org/dhis2/data/dhislogic/DhisProgramUtils.kt
@@ -161,7 +161,7 @@ class DhisProgramUtils @Inject constructor(val d2: D2) {
             ProgramType.WITH_REGISTRATION -> {
                 program.trackedEntityType()?.displayName() ?: defaultTrackerLabel
             }
-            ProgramType.WITHOUT_REGISTRATION -> program.eventLabel() ?: defaultEventLabel
+            ProgramType.WITHOUT_REGISTRATION -> defaultEventLabel
             null -> ""
         }
     }
--- a/app/src/main/java/org/dhis2/usescases/enrollment/EnrollmentActivity.kt
+++ b/app/src/main/java/org/dhis2/usescases/enrollment/EnrollmentActivity.kt
@@ -16,6 +16,7 @@ import org.dhis2.commons.Constants.TEI_UID
 import org.dhis2.commons.data.TeiAttributesInfo
 import org.dhis2.commons.dialogs.imagedetail.ImageDetailActivity
 import org.dhis2.commons.featureconfig.data.FeatureConfigRepository
+import org.dhis2.commons.resources.EventResourcesProvider
 import org.dhis2.commons.resources.ResourceManager
 import org.dhis2.databinding.EnrollmentActivityBinding
 import org.dhis2.form.data.GeometryController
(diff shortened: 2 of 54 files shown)


app/src/main/java/org/dhis2/data/dhislogic/DhisProgramUtils.kt
```diff
@@ -161,7 +161,7 @@ class DhisProgramUtils @Inject constructor(val d2: D2) {
             ProgramType.WITH_REGISTRATION -> {
                 program.trackedEntityType()?.displayName() ?: defaultTrackerLabel
             }
-            ProgramType.WITHOUT_REGISTRATION -> defaultEventLabel
+            ProgramType.WITHOUT_REGISTRATION -> program.eventLabel() ?: defaultEventLabel
```
andresmr, line 164: isn't this method deprecated? aren't they changing to `program.displayEventLabel()`?

app/src/main/java/org/dhis2/usescases/enrollment/EnrollmentActivity.kt
```diff
@@ -357,7 +357,13 @@ class EnrollmentActivity : ActivityGlobalAbstract(), EnrollmentView {
 
     override fun showDateEditionWarning() {
         val dialog = MaterialAlertDialogBuilder(this, R.style.DhisMaterialDialog)
-            .setMessage(R.string.enrollment_date_edition_warning)
+            .setMessage(
+                resourceManager.formatWithProgramEventLabel(
+                    R.string.enrollment_date_edition_warning_event_label,
+                    presenter.getProgram()?.uid(),
```
andresmr, line 363: what do you think about send the program uid in the `showDateEditionWarning` as a parameter instead of using the presenter as a getter?

app/src/main/java/org/dhis2/usescases/programStageSelection/ProgramStageSelectionActivity.kt
```diff
@@ -26,6 +27,9 @@ class ProgramStageSelectionActivity : ActivityGlobalAbstract(), ProgramStageSele
     @Inject
     lateinit var presenter: ProgramStageSelectionPresenter
 
+    @Inject
+    lateinit var resourceManager: ResourceManager
```
andresmr, line 31: What do you think about injecting the `EventResourceProvider` to reduce duplicated code?

commons/src/main/java/org/dhis2/commons/resources/ResourceManager.kt
```diff
@@ -109,24 +109,58 @@ class ResourceManager(
...
             .programStages().uid(programStageUid)
-            .blockingGet()?.eventLabel()
+            .blockingGet()?.displayEventLabel()
     } catch (e: Exception) {
         null
-    } ?: getPlural(R.plurals.event_label, quantity)
+    } ?: programEventLabel(programUid, quantity)
 
-    fun formatWithEventLabel(
+    fun formatWithProgramStageEventLabel(
```
andresmr, line 124: If we add theses specific labels in the resource manager it will grow too much.
What about create a `EventResourcersProvider` to provide all resources of the event, and in the future the same for program etc... with this we develop the code once per label injecting the `resourcesManager` to get the strings, with this we leave resourcesManager just to get resources and we develop one provider per scope. What do you think?

andresmr (changes requested):

app/src/main/java/org/dhis2/data/dhislogic/DhisProgramUtils.kt
mmmateos: You are rigth, in fact this doesn't came here, I forgot to remove it. Thanks

app/src/main/java/org/dhis2/usescases/enrollment/EnrollmentActivity.kt
mmmateos: Yes, better way! thanks

commons/src/main/java/org/dhis2/commons/resources/ResourceManager.kt
mmmateos: Yes, definitely a better approach. I will try to refactor this, thanks for the idea!

andresmr (commented):
Great changes!, I have a last comment/doubt, do you think it would be better to merge `EventResourcesProvider` and `EventDetailResourcesProvider` having just one provider class with methods like: provideEventDate, provideLabel etc... so that the strings keeps inside the provider and doesn't have to be given by activities and so on?.

xavimolloy (approved):

mmmateos:
@andresmr Not sure, as I understand EventDetailsResourceProvider gives all resources for the Event details context, but EventResourcesProviders is giving the resource for the word event along the app. We can discuss but I understand it as separates things.

PR #3786, opened 2024-08-29
chore: [ANDROAPP-6341] Use displayContent with fallback to content in rule actions

taridepaco:
Show translation of ProgramRuleAction, like Warnings and Errors non-functional, on the  Android app.

[ jira issue ](https://dhis2.atlassian.net/browse/ANDROAPP-6341)

### DIFF
--- a/dhis2-mobile-program-rules/src/main/java/org/dhis2/mobileProgramRules/RuleEngineExtensions.kt
+++ b/dhis2-mobile-program-rules/src/main/java/org/dhis2/mobileProgramRules/RuleEngineExtensions.kt
@@ -111,7 +111,7 @@ fun ProgramRuleAction.toRuleEngineObject(): RuleAction {
                 values = mutableMapOf(
                     Pair("field", field),
                 ).also { map ->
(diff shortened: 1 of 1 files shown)


xavimolloy (approved):

dhis2-mobile-program-rules/src/main/java/org/dhis2/mobileProgramRules/RuleEngineExtensions.kt
```diff
@@ -111,7 +111,7 @@ fun ProgramRuleAction.toRuleEngineObject(): RuleAction {
                 values = mutableMapOf(
                     Pair("field", field),
                 ).also { map ->
-                    content()?.let { map["content"] = it }
+                    (displayContent() ?: content())?.let { map["content"] = it }
```
vgarciabnz, line 114: This property, is it evaluated for all the action types? If so, we could evaluate only once at the beginning of the method (out of the `when`, next to `field`).

vgarciabnz (approved):

PR #3796, opened 2024-09-17
Display new schedule dialog UI when creating schedule

msasikanth:


andresmr (changes requested):

app/src/main/java/org/dhis2/usescases/teiDashboard/dashboardfragments/teidata/TEIDataPresenter.kt
```diff
@@ -359,13 +359,20 @@ class TEIDataPresenter(
...
+        stage: ProgramStage?,
+    ) {
         if (stage != null) {
             when (eventCreationType) {
                 EventCreationType.ADDNEW -> programUid?.let { program ->
                     checkOrgUnitCount(program, stage.uid())
                 }
 
+                EventCreationType.SCHEDULE -> {
+                    view.displayScheduleEvent(showYesNoOptions = false)
```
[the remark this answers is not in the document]
msasikanth: Here we have the possibility to get one or null program stage. So if we already have the program stage, we can simply hide the program selection from the scheduling dialog, or else if it's null, we display it. Is that what you are expecting?

xavimolloy (approved):

app/src/main/java/org/dhis2/usescases/teiDashboard/dashboardfragments/teidata/TEIDataPresenter.kt
```diff
@@ -359,17 +362,30 @@ class TEIDataPresenter(
...
+                        programStage = stage,
+                        showYesNoOptions = false,
+                    )
+                }
+
                 else -> view.goToEventInitial(eventCreationType, stage)
             }
         } else {
-            createEventInEnrollment(eventCreationType)
+            view.displayScheduleEvent(
```
andresmr, line 385: The method that you changed `createEventInEnrollment` it's only used in a test. What do you think about removing it?.
Also this method was navigating to `ProgramStageSelectionActivity` which ( I think is no used anymore). Could you check and if so, remove it?
Thanks!.

andresmr (changes requested):

app/src/main/java/org/dhis2/usescases/teiDashboard/dashboardfragments/teidata/TEIDataPresenter.kt
msasikanth: Done ✅

PR #3801, opened 2024-09-23
feat: [ANDROAPP-6229] Display new schedule dialog UI when creating/entering schedule

msasikanth:


msasikanth:
@andresmr I have noticed an edge case (?) scenario in this flow. When user clicks on enter event, we only show due date option now and no report date. When they click on enter event, there is no longer a report date pre-filled there. If we back press, and click on "not now", the app is crashing when loading the TEI fragment?

xavimolloy (approved):

andresmr (approved):

PR #3806, opened 2024-10-01
ANDROAPP-6432-search-screen-navigation-bar

msasikanth:


app/src/main/java/org/dhis2/usescases/searchTrackEntity/ui/SearchScreenConfigurator.kt
```diff
@@ -6,6 +6,8 @@ import androidx.constraintlayout.widget.ConstraintSet
 import org.dhis2.R
 import org.dhis2.bindings.display
 import org.dhis2.bindings.dp
+import org.dhis2.commons.animations.hide
+import org.dhis2.commons.animations.show
```
andresmr, line 10: do we need these imports?

msasikanth: Yep, since these are extension functions on `View`

andresmr (approved):

xavimolloy (approved):

app/src/main/java/org/dhis2/usescases/searchTrackEntity/SearchTEActivity.kt
```diff
@@ -0,0 +1,789 @@
...
+        viewModel.updateActiveFilters(totalFilters > 0)
+        viewModel.refreshData()
+    }
+
+    private fun initSearchParameters() {
+        initSearchScreen(
+            binding.searchContainer,
+            viewModel,
+            initialProgram,
+            tEType!!,
```
mmmateos, line 336: Couldn't this lead to a crash if there is no teType?

msasikanth: It doesn't look like it's going to be a null value, the reason it's a nullable is because the platform `intent.getStringExtra` is marked as nullable. But in our scenario we always pass it when creating this activity.

To handle this, I have converted the variable to a `lateinit` and set the default intent extra in case it's null to empty string. At least we don't have to force cast to non-null now.

msasikanth: I have also cleaned up some of the other non-null assertions

PR #3811, opened 2024-10-04
ANDROAPP-6431-event-programs-navigation-bar

msasikanth:


app/src/main/java/org/dhis2/usescases/programEventDetail/ProgramEventDetailViewModel.kt
```diff
@@ -8,16 +17,24 @@ import androidx.lifecycle.viewModelScope
...
 import org.dhis2.commons.viewmodel.DispatcherProvider
 import org.dhis2.maps.layer.basemaps.BaseMapStyle
 import org.dhis2.maps.usecases.MapStyleConfiguration
+import org.dhis2.tracker.NavigationBarUIState
 import org.dhis2.usescases.programEventDetail.usecase.CreateEventUseCase
+import org.dhis2.utils.customviews.navigationbar.NavigationPage
+import org.dhis2.utils.customviews.navigationbar.NavigationPageConfigurator
+import org.hisp.dhis.mobile.ui.designsystem.component.navigationBar.NavigationBarItem
 
 class ProgramEventDetailViewModel(
```
ferdyrod, line 31: This is a good opportunity to add tests to this view model

app/src/main/java/org/dhis2/usescases/programEventDetail/ProgramEventDetailViewModel.kt
```diff
@@ -43,6 +60,63 @@ class ProgramEventDetailViewModel(
     val shouldNavigateToEventDetails: SharedFlow<String>
         get() = _shouldNavigateToEventDetails
 
+    private val _navigationBarUIState = mutableStateOf(NavigationBarUIState<NavigationPage>())
+    val navigationBarUIState: State<NavigationBarUIState<NavigationPage>> = _navigationBarUIState
+
+    init {
+        viewModelScope.launch { loadBottomBarItems() }
+    }
```
ferdyrod, line 68: I would recommend not to do any initialization in a viewModel's `init{}` because it can cause weird behavior on config changes and also, it will be harder to test.

Maybe you could use flows and emit the changes of the bar's items

app/src/main/java/org/dhis2/usescases/programEventDetail/ProgramEventDetailViewModel.kt
msasikanth: Yep, let me add a test for this change

xavimolloy (approved):

ferdyrod (approved):

PR #3832, opened 2024-10-22
fix: [ANDROAPP-6520] save complete event to Preferences from form module

xavimolloy:
## Description

## Solution description
## Covered unit test cases
- [ ] Smartphone Emulator
- [ ] Tablet Emulator
- [ ] Smartphone
- [ ] Tablet
- [ ] 4.4
- [ ] 5.X - 6.X
- [ ] 7.X
- [ ] 8.X
- [ ] 9.X - 10.X
- [ ] 11.X - 13.X
- [ ] Other
## Checklist

### DIFF
--- a/form/src/main/java/org/dhis2/form/ui/FormViewModel.kt
+++ b/form/src/main/java/org/dhis2/form/ui/FormViewModel.kt
@@ -6,8 +6,6 @@ import androidx.lifecycle.LiveData
 import androidx.lifecycle.MutableLiveData
 import androidx.lifecycle.ViewModel
 import androidx.lifecycle.viewModelScope
-import com.google.gson.Gson
-import com.google.gson.reflect.TypeToken
 import kotlinx.coroutines.async
 import kotlinx.coroutines.channels.BufferOverflow
 import kotlinx.coroutines.channels.Channel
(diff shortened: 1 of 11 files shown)


ferdyrod (approved):

Balcan (approved):

form/src/main/java/org/dhis2/form/ui/FormViewModel.kt
```diff
@@ -766,6 +767,7 @@ class FormViewModel(
             try {
                 async(dispatcher.io()) {
                     repository.completeEvent()
+                    preferenceProvider.setValue(Preference.PREF_COMPLETED_EVENT, repository.getRecordUid())
```
andresmr, line 770: Why do we need to expose a value from valueStore(_data_) to ViewModel(_ui_) to store it in preferenceProvdier(_data_) ?. I was thinking for example... if it were a Use Case (ie: CompleteEvent), it can complete in SDK and then update preferences. As we don't want to do any refactor y the code freeze, I would set this reponsibility to the Repository.
What do you think to inject preferenceProvider in FormRepositoryImpl and in fun complete() do the valueStore complete plus the updating preferences?

xavimolloy: You are right, I wanted to avoid injecting the PreferenceProvider in another place as we already had access to it from the viewmodel, although we didn't have access to the record uid. I will also remove the PreferenceProvider calls from the viewmodel to the repositoryso that it's all in the same place.

andresmr (approved):

PR #3857, opened 2024-10-31
fix: [ANDROAPP-6605] Use Network provider for location services

xavimolloy:
## Description

## Solution description
## Covered unit test cases
- [ ] Smartphone Emulator
- [ ] Tablet Emulator
- [ ] Smartphone
- [ ] Tablet
- [ ] 4.4
- [ ] 5.X - 6.X
- [ ] 7.X
- [ ] 8.X
- [ ] 9.X - 10.X
- [ ] 11.X - 13.X
- [ ] Other
## Checklist

### DIFF
--- a/commons/src/main/java/org/dhis2/commons/locationprovider/LocationProviderImpl.kt
+++ b/commons/src/main/java/org/dhis2/commons/locationprovider/LocationProviderImpl.kt
@@ -8,8 +8,8 @@ import android.content.pm.PackageManager
 import android.location.Location
 import android.location.LocationListener
 import android.location.LocationManager
-import android.os.Bundle
 import androidx.core.app.ActivityCompat
+import androidx.core.location.LocationListenerCompat
 
 private const val FUSED_LOCATION_PROVIDER = "fused"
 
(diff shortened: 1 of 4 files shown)


andresmr (approved):

commons/src/main/java/org/dhis2/commons/locationprovider/LocationProviderImpl.kt
```diff
@@ -67,15 +67,10 @@ open class LocationProviderImpl(val context: Context) : LocationProvider {
...
                 }
-
-                @Deprecated("Deprecated in Java")
-                override fun onStatusChanged(provider: String?, status: Int, extras: Bundle?) {
-                    // Need implementation for compatibility
-                }
             }
 
             locationManager.requestLocationUpdates(
-                LocationManager.GPS_PROVIDER,
```
Balcan, line 78: We can't do this. NETWORK provider uses wifi access points and cell towers which won't be available in certain locations. I think the best would be combining both, creating two locationRequests, one with the network provider which should be faster and other one with the gps.

Balcan (changes requested):

commons/src/main/java/org/dhis2/commons/locationprovider/LocationProviderImpl.kt
xavimolloy: I have updated the PR including the call to the GPS provider.

andresmr (approved):

Balcan (approved):

PR #3863, opened 2024-11-05
rename displayName property on NominatimLocation

andresmr:
Remove code smell


xavimolloy (approved):

Balcan (changes requested):
You can't rename the variable. The json parser needs to have the same exact key

Balcan (approved):

PR #3875, opened 2024-11-12
fix: [ANDROAPP-6622] Add progress to search location

Balcan:
## Description

## Solution description
## Covered unit test cases
- [ ] Smartphone Emulator
- [ ] Tablet Emulator
- [ ] Smartphone
- [ ] Tablet
- [ ] 4.4
- [ ] 5.X - 6.X
- [ ] 7.X
- [ ] 8.X
- [ ] 9.X - 10.X
- [ ] 11.X - 13.X
- [ ] Other
## Checklist

### DIFF
--- a/commons/build.gradle.kts
+++ b/commons/build.gradle.kts
@@ -52,7 +52,7 @@ android {
     }
 
     configurations.all {
-        resolutionStrategy.cacheDynamicVersionsFor(0, TimeUnit.SECONDS)
+        resolutionStrategy.cacheChangingModulesFor(0, TimeUnit.SECONDS)
     }
 }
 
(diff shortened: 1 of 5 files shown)


commons/build.gradle.kts
```diff
@@ -52,7 +52,7 @@ android {
     }
 
     configurations.all {
-        resolutionStrategy.cacheDynamicVersionsFor(0, TimeUnit.SECONDS)
+        resolutionStrategy.cacheChangingModulesFor(0, TimeUnit.SECONDS)
```
andresmr, line 55: which is the impact of this?

Balcan: Dynamic version is when we have a version specified with "+". Which we do not have. Changing version is when we have a dependency with the isChanging = true, of which we have 2, the sdk and the mobile ui. This way the dependency won't be cached and any new change will be downloaded. It was not working until now, that's why we had problems with some snapshot versions.

andresmr (approved):

PR #3887, opened 2024-11-18
Use GH API to push commits in gihtub actions

vgarciabnz:
## Description
This PR uses the [DHIS2-bot](https://github.com/dhis2-bot) to create the commit via GH API. In this way, Github automatically signed the commits. 

If makes use of a plugin called [gh-commit](https://github.com/flex-development/gh-commit), which wraps the communication with the GH API via GraphQL API. 

There is a limitation with protected branches: it seems that is not possible to push to protected branches via GraphQL (or GH API) even if the user is included in the Bypass list. As a workaround, the action creates an auxiliary branch to create the commit via API (so the commit is signed) and then it creates the `release/...` branch from this auxiliary branch. And deletes the auxiliary branch. This workaround could be avoided if we manage to push commits to protected branches via GH API.

[ANDROAPP-6643](https://dhis2.atlassian.net/browse/ANDROAPP-6643)

## Covered unit test cases
Manually tested by triggering GH actions.

## Checklist
- [X] My code follows the style guidelines of this project
- [X] I have performed a self-review of my own code

[ANDROAPP-6643]: https://dhis2.atlassian.net/browse/ANDROAPP-6643?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/workflows/release-generate-notes.yml
+++ b/.github/workflows/release-generate-notes.yml
@@ -22,19 +22,16 @@ jobs:
     steps:
     
       - uses: actions/checkout@v3
+        with:
+          token: ${{ secrets.DHIS2_BOT_GITHUB_TOKEN }}
+
       # Generate github release notes
       - name: Generate release notes
         working-directory: ./scripts
         run: python3 generateReleaseNotes.py
 
-      - name: setup git config
-        run: |
-          # setup the username and email.
-          git config user.name "GitHub Actions Bot"
-          git config user.email "<android@dhis2.org>"
-
       - name: Commit changes
-        run: |
-          # Commit and push
-          git commit -am "Update release notes"
-          git push
+        uses: flex-development/gh-commit@1.0.0
+        with:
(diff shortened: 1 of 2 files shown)


.github/workflows/release-generate-notes.yml
```diff
@@ -22,19 +22,16 @@ jobs:
...
-          # setup the username and email.
-          git config user.name "GitHub Actions Bot"
-          git config user.email "<android@dhis2.org>"
-
       - name: Commit changes
-        run: |
-          # Commit and push
-          git commit -am "Update release notes"
-          git push
+        uses: flex-development/gh-commit@1.0.0
```
andresmr, line 34: does this also push?

andresmr (approved):

Balcan (approved):

PR #3905, opened 2024-11-27
fix: [ANDROAPP-6579] set dropdownmenu max width to 70% of screen size

Balcan:
## Description

## Solution description
## Covered unit test cases
- [ ] Smartphone Emulator
- [ ] Tablet Emulator
- [ ] Smartphone
- [ ] Tablet
- [ ] 4.4
- [ ] 5.X - 6.X
- [ ] 7.X
- [ ] 8.X
- [ ] 9.X - 10.X
- [ ] 11.X - 13.X
- [ ] Other
## Checklist

### DIFF
--- a/app/src/main/java/org/dhis2/utils/customviews/MoreMenuView.kt
+++ b/app/src/main/java/org/dhis2/utils/customviews/MoreMenuView.kt
@@ -8,9 +8,11 @@ import androidx.compose.material3.Icon
 import androidx.compose.runtime.Composable
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.unit.Dp
+import androidx.compose.ui.unit.DpOffset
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/dhis2/utils/customviews/MoreMenuView.kt
```diff
@@ -39,3 +43,8 @@ fun <T> MoreOptionsWithDropDownMenuButton(
         onItemClick(itemId)
     }
 }
+
+inline val Double.dw: Dp
```
andresmr, line 47: why do we need this function? should we extract it to some place to reuse it in other places?

Balcan: It is just a way to calculate the percentage for the max width. I don't think it needs to be on the mobile-ui, but could be useful to move it to commons.

PR #3980, opened 2025-02-05
Enable support to login with two factor code

ifoche:
## Description

No Jira issue for this. Just an external contribution from EyeSeeTea.

## Solution description
This PR enables the use of 2FA in servers above v41. It depends on [this PR SDK side](https://github.com/dhis2/dhis2-android-sdk/pull/2311).

When a user activates 2FA the app will detect the failure produced by a login attempt on a 2FA-enabled user and will add a field asking for 2FA code to that user. 

The 2FA requires the cookie to be present after that login. So when the login expires (this is e.g. when its been a while since the user didn't login) the sync attemps will fail. It is important to note that, effectively, this means that when a user activates its 2FA, the user needs to be trained in the different from a sync perspective, as the app will only sync its metadata/data after a login is done. It's then important that the user knows that without a login, the data that was not synced won't be synced to the server before the next login.

## Covered unit test cases
- [x] Smartphone Emulator
- [ ] Tablet Emulator
- [x] Smartphone
- [x] Tablet
- [ ] 4.4
- [ ] 5.X - 6.X
- [ ] 7.X
- [ ] 8.X
- [x] 9.X - 10.X
- [x] 11.X - 13.X
- [x] Other
## Checklist

### DIFF
--- a/app/src/main/java/org/dhis2/data/server/OpenIdSession.kt
+++ b/app/src/main/java/org/dhis2/data/server/OpenIdSession.kt
@@ -25,6 +25,7 @@ class OpenIdSession(
     enum class LogOutReason {
         OPEN_ID,
         DISABLED_ACCOUNT,
+        UNAUTHORIZED,
     }
 
     fun setSessionCallback(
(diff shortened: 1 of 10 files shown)


andresmr:
Hi @ifoche we are closing the PR for now, we'll check for future versions. Thanks for you contribution!

PR #4060, opened 2025-03-19
fix: [ANDROAPP-6903] update org unit only if any selected

andresmr:
## Description
Save org unit only if selected, not with dismiss

[ jira issue ](https://dhis2.atlassian.net/browse/ANDROAPP-903)


Balcan (approved):

xavimolloy (commented):
can you fix the conflicts?

PR #4083, opened 2025-03-27
chore: [ANDROAPP-6949] add PR size check to Jenkins pipeline

andresmr:
## Description
Limit the PR size on jenkins pipeline

[ jira issue ](https://dhis2.atlassian.net/browse/ANDROAPP-6949)


scripts/check_pr_size.sh
```diff
@@ -0,0 +1,38 @@
+#!/bin/bash
+
+MAX_LINES=400
```
Balcan, line 3: It can be usefull to handle this value from within Github environment variables. Thay way, if we need to fine tune it we don't need to change the script. What do you think?

Balcan (approved):

xavimolloy (approved):

PR #4240, opened 2025-07-08
feat: [ANDROAPP-6843] Implement custom intent launch and data recovery

xavimolloy:
…unching and data recovery

## Description

Summarize the change and link
the [JIRA issue](https://dhis2.atlassian.net/browse/ANDROAPP-<ISSUE-NUMBER>).

## Solution Description

Briefly describe how the issue is resolved.

## Title Format

`<type>: [JIRA-ISSUE] Description`

- `<type>` Follow [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) (e.g., `fix`, `feat`).
- Add the **JIRA issue number** in `[ANDROAPP-XXXX]`.
- Include a brief description.
- Append `[skip ci]` to skip CI if needed.
- Append `[skip size]` to skip PR size limitation.


form/src/main/java/org/dhis2/form/data/DataEntryBaseRepository.kt
```diff
@@ -28,6 +28,9 @@ abstract class DataEntryBaseRepository(
     private val metadataIconProvider: MetadataIconProvider,
 ) : DataEntryRepository {
 
+    protected val formConfiguration: FormBaseConfiguration
```
Balcan, line 31: we already have this variable in the constructor. Can't we use it instead of creating a duplicate?

form/src/main/java/org/dhis2/form/data/EventRepository.kt
```diff
@@ -67,6 +67,7 @@ class EventRepository(
         EventPeriodRepository(d2),
     )
 
+    private val config = formConfiguration
```
Balcan, line 70: Same here

form/src/main/java/org/dhis2/form/data/DataEntryBaseRepository.kt
xavimolloy: I thought there could be a better way of doing this, thanks for the help, I have updated the code

andresmr (approved):

PR #4260, opened 2025-07-14
Update dependencies only with major and minor versions

Balcan:
## Description

Summarize the change and link
the [JIRA issue](https://dhis2.atlassian.net/browse/ANDROAPP-<ISSUE-NUMBER>).

## Solution Description

Briefly describe how the issue is resolved.

## Title Format

`<type>: [JIRA-ISSUE] Description`

- `<type>` Follow [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) (e.g., `fix`, `feat`).
- Add the **JIRA issue number** in `[ANDROAPP-XXXX]`.
- Include a brief description.
- Append `[skip ci]` to skip CI if needed.
- Append `[skip size]` to skip PR size limitation.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -14,6 +14,12 @@ updates:
     open-pull-requests-limit: 5
     commit-message:
       prefix: "deps"
+    ignore:
+      - dependency-name: "com.squareup.okhttp3:*"
+      - dependency-name: "*"
+        update-types:
+          - "version-update:semver-major"
+          - "version-update:semver-minor"
     groups:
       gradle-updates:
         patterns:


andresmr (approved):

.github/dependabot.yml
```diff
@@ -14,6 +14,12 @@ updates:
     open-pull-requests-limit: 5
     commit-message:
       prefix: "deps"
+    ignore:
+      - dependency-name: "com.squareup.okhttp3:*"
```
andresmr, line 18: why we don't want to update OKHTTP?

Balcan: We get different version from the sdk which caused issues in the past. So it is better to update it manually when the sdk is updated.

PR #4298, opened 2025-08-05
feat: [ANDROAPP-7164] Evaluate custom intent request parameters

vgarciabnz:
## Description

This is a draft PR for [ANDROAPP-7164](https://dhis2.atlassian.net/browse/ANDROAPP-7164) just to discuss about the proposed solution. This PR shouldn't be merged as it is.

## Solution Description

It uses the DHIS2 Android SDK to evaluate the custom intent request parameters. The purpose of this evaluation is to be able to use variables and different value types in the parameters.

The parameters are evaluated when the form is built, which means that they are not re-evaluated when the form is been filled in. This might be a problem for certain variables: for example, if we add the variable `V{org_unit_id}` and the orgunit changes after the form is built, the parameter will still have the value from the previous orgunit. 

Ideally, this evaluation should be done right before launching the intent. Do you think it would be possible?

[ANDROAPP-7164]: https://dhis2.atlassian.net/browse/ANDROAPP-7164?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/commonskmm/src/commonMain/kotlin/org/dhis2/mobile/commons/model/CustomIntentModel.kt
+++ b/commonskmm/src/commonMain/kotlin/org/dhis2/mobile/commons/model/CustomIntentModel.kt
@@ -10,7 +10,7 @@ data class CustomIntentModel(
 
 data class CustomIntentRequestArgumentModel(
     val key: String,
-    val value: String,
+    val value: Any,
 )
 
 data class CustomIntentResponseDataModel(
--- a/form/src/main/java/org/dhis2/form/ui/provider/inputfield/CustomIntentProvider.kt
+++ b/form/src/main/java/org/dhis2/form/ui/provider/inputfield/CustomIntentProvider.kt
@@ -112,7 +112,15 @@ fun ProvideCustomIntentInput(
 fun mapIntentData(customIntent: CustomIntentModel): Intent {
     return Intent(customIntent.packageName).apply {
         customIntent.customIntentRequest.forEach { argument ->
-            putExtra(argument.key, argument.value)
+            when (val value = argument.value) {
+                is String -> putExtra(argument.key, value)
+                is Int -> putExtra(argument.key, value)
+                is Double -> putExtra(argument.key, value)
+                is Long -> putExtra(argument.key, value)
+                is Short -> putExtra(argument.key, value)
+                is Boolean -> putExtra(argument.key, value)
+                else -> putExtra(argument.key, value.toString())
+            }
         }
     }
 }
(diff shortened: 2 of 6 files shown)


andresmr (approved):

commonskmm/src/commonMain/kotlin/org/dhis2/mobile/commons/model/CustomIntentModel.kt
```diff
@@ -10,7 +10,7 @@ data class CustomIntentModel(
 
 data class CustomIntentRequestArgumentModel(
     val key: String,
-    val value: String,
+    val value: Any,
```
Balcan, line 13: Wouldn't it be better to use a wildcard instead of Any?

vgarciabnz: What do you mean? I think that having `val value: *` is not valid. How would it be?

form/src/main/java/org/dhis2/form/ui/provider/inputfield/CustomIntentProvider.kt
```diff
@@ -112,7 +112,15 @@ fun ProvideCustomIntentInput(
 fun mapIntentData(customIntent: CustomIntentModel): Intent {
     return Intent(customIntent.packageName).apply {
         customIntent.customIntentRequest.forEach { argument ->
-            putExtra(argument.key, argument.value)
+            when (val value = argument.value) {
+                is String -> putExtra(argument.key, value)
```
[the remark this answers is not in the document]
vgarciabnz: Well, I don't know... These parameters are the result of a evaluation in the expression-parser, which might basically return a String, a Boolean or a Number (Integer, Short,...). For example, it won't return an object or a list of objects. I think it won't be very easy to align them

commonskmm/src/commonMain/kotlin/org/dhis2/mobile/commons/model/CustomIntentModel.kt
vgarciabnz: If you do so, then I guess the model will have to be:
```
data class CustomIntentModel(
  ...
  val customIntentRequest: List<CustomIntentRequestArgumentModel<*>>,
)
```
Each request argument might have a different type.

That is valid as well, but I am not sure about the advantage. If you usually do it this way in the app, that's fine, up to you.

andresmr: I have been looking and it seems that `value: Any` is not type safe so you need to perform manual casting what can lead to runtime errors. `value: T` is type safe in compilation time but adds complexity when serializing or deserializing.
As we don't need to deserialize the object, because we map it, I think that it's better to use `value: T`

PR #4318, opened 2025-08-21
fix: [ANDROAPP-6844] Adapt Search to use list of parameters

xavimolloy:
## Description


app/src/main/java/org/dhis2/usescases/searchTrackEntity/SearchRepositoryImpl.java
```diff
@@ -225,13 +225,23 @@ public TrackedEntitySearchCollectionRepository getFilteredRepository(SearchParam
...
-                } else if (dataValue.contains("_os_")) {
-                    dataValue = dataValue.split("_os_")[1];
-                    trackedEntityInstanceQuery = trackedEntityInstanceQuery.byFilter(dataId).eq(dataValue);
-                } else
-                    trackedEntityInstanceQuery = trackedEntityInstanceQuery.byFilter(dataId).like(dataValue);
+                if(dataValues != null && dataValues.size()>1) {
+                    trackedEntityInstanceQuery = trackedEntityInstanceQuery.byFilter(dataId).in(dataValues);
+
+                } else {
+                    assert dataValues != null;
```
andresmr, line 232: you are asserting twice that `dataVaues != null`, here and in [line 228](https://github.com/dhis2/dhis2-android-capture-app/pull/4318/files#diff-ea895a67063b379e5a1b677889df995f59fd5333c349527593ee3a5a5007b03dR228) , maybe we can simplify this block.

app/src/main/java/org/dhis2/usescases/searchTrackEntity/SearchRepositoryImpl.java
```diff
@@ -225,13 +225,23 @@ public TrackedEntitySearchCollectionRepository getFilteredRepository(SearchParam
...
+                    if(dataValues.size() == 1) {
+                        String dataValue = dataValues.get(0);
+                        if (isUnique || isOptionSet) {
+                            trackedEntityInstanceQuery = trackedEntityInstanceQuery.byFilter(dataId).eq(dataValue);
+                        } else if (dataValue.contains("_os_")) {
```
andresmr, line 237: why don't you extract this hard coded string to a REGEX constant?

app/src/main/java/org/dhis2/usescases/searchTrackEntity/SearchRepositoryImpl.java
```diff
@@ -225,13 +225,23 @@ public TrackedEntitySearchCollectionRepository getFilteredRepository(SearchParam
...
+                            dataValue = dataValue.split("_os_")[1];
+                            trackedEntityInstanceQuery = trackedEntityInstanceQuery.byFilter(dataId).eq(dataValue);
+                        } else
+                            trackedEntityInstanceQuery = trackedEntityInstanceQuery.byFilter(dataId).like(dataValue);
+                    }
+                }
+
```
andresmr, line 244: All this block is difficult to understand, as it is a legacy class, maybe is not the moment to refactor it now, but why don't you add some line comments to explain what is doing on each condition?

app/src/main/java/org/dhis2/usescases/searchTrackEntity/SearchTEExtra.kt
```diff
@@ -25,11 +26,12 @@ fun SearchTEActivity.queryDataExtra(savedInstanceState: Bundle?): Map<String, St
                     ?.toList() ?: emptyList()
             if (attributes.size != values.size) return emptyMap()
             attributes.mapIndexed { index, attributeUid ->
-                attributeUid to values[index]
+                attributeUid to listOf(values[index])
             }.toMap()
         }
         savedInstanceState.containsKey(Constants.QUERY_DATA) -> {
-            savedInstanceState.getSerializable(Constants.QUERY_DATA) as Map<String, String>
+            @Suppress("UNCHECKED_CAST")
```
andresmr, line 33: Do you think that is better set the annotation here or at class level?

app/src/main/java/org/dhis2/usescases/searchTrackEntity/SearchTEExtra.kt
```diff
@@ -25,11 +26,12 @@ fun SearchTEActivity.queryDataExtra(savedInstanceState: Bundle?): Map<String, St
...
+            savedInstanceState.getSerializable(Constants.QUERY_DATA) as Map<String, List<String>>
```
andresmr, line 34: `getSerializable` is deprecated, can you find another alternative?. Maybe using Parcelable

app/src/test/java/org/dhis2/usescases/searchTrackEntity/SearchTEIViewModelTest.kt
andresmr, line 1: would be possible to add tests to cover the possible scenarios that you add? like receiving a list of query parameters for one key, receiving a single one or even receiving more than one list of query parameters per key.

andresmr (changes requested):
The PR is clear to understand and easy to follow. **Congratulations**!.
I added some comments oriented to increase readability, test coverage and some doubts I had understanding the code.

app/src/main/java/org/dhis2/usescases/searchTrackEntity/SearchRepositoryImpl.java
xavimolloy: I will add some clarification comments

PR #4334, opened 2025-08-26
feat: [ANDROAPP-6844] [CUSTOM INTENTS SEARCH III]  implement call and ui rendering in search

xavimolloy:
## Description

(https://dhis2.atlassian.net/jira/software/c/projects/ANDROAPP/boards/113?selectedIssue=ANDROAPP-6844)

The implementation for the UI field and the SDK request to fetch the custom intents for the search flow.


commonskmm/src/commonMain/kotlin/org/dhis2/mobile/commons/model/CustomIntentModel.kt
```diff
@@ -27,3 +27,8 @@ enum class CustomIntentResponseExtraType {
     OBJECT,
     LIST_OF_OBJECTS,
 }
+
+enum class CustomIntentActionTypeModel {
```
Balcan, line 31: Not to be changed, but I think we should be using sealed interfaces with data objects instead of enums. Maybe in the future we want to expand the SEARCH with a set of parameters or something like that and having data classes/objets are easier to update. Also, for multiplatform I think enums could be problematic.

Balcan (approved):

xavimolloy:
Ui tests are passing locally but failing in browserstack

PR #4399, opened 2025-10-06
fix [ANDROAPP-7234] Allow to search using comma

xavimolloy:
## Description

 [JIRA issue](https://dhis2.atlassian.net/jira/software/c/projects/ANDROAPP/boards/113?selectedIssue=ANDROAPP-7234).

With the new custom intent implementation in the Search TEI section, we could not search teiAttributes with a value with a comma in it, in order to adapt this, a new method has been added to the custom intent repository that checks whether that TEI attribute is linked to a custom intent that can return a list of values and then will search by it, if not, it will search with the plain text that was introduced in the search field

### DIFF
--- a/app/src/main/java/org/dhis2/usescases/searchTrackEntity/SearchRepositoryImpl.java
+++ b/app/src/main/java/org/dhis2/usescases/searchTrackEntity/SearchRepositoryImpl.java
@@ -27,6 +27,8 @@ import org.dhis2.data.sorting.SearchSortingValueSetter;
 import org.dhis2.metadata.usecases.FileResourceConfiguration;
 import org.dhis2.metadata.usecases.ProgramConfiguration;
 import org.dhis2.metadata.usecases.TrackedEntityInstanceConfiguration;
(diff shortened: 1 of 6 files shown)


Balcan (approved):

app/src/main/java/org/dhis2/usescases/searchTrackEntity/SearchRepositoryImpl.java
```diff
@@ -122,6 +124,7 @@ public class SearchRepositoryImpl implements SearchRepository {
 
     private final MetadataIconProvider metadataIconProvider;
     private final ProfilePictureProvider profilePictureProvider;
+    private CustomIntentRepository customIntentRepository;
```
andresmr, line 127: I think that repositories shouldn't be dependent on other repositories, domain logic (use cases, interactors...) should handle that, anyway given that this code need to be refactored, if you consider is the best solution for now, it's OK for me 😄 .

andresmr (approved):

PR #4527, opened 2025-11-28
fix: [ANDROAPP-6870] remove useCodeForOptionSet logic [skip size]

Balcan:
## Description

### DIFF
--- a/app/src/androidTest/java/org/dhis2/usescases/programevent/ProgramEventTest.kt
+++ b/app/src/androidTest/java/org/dhis2/usescases/programevent/ProgramEventTest.kt
@@ -6,7 +6,6 @@ import androidx.compose.ui.test.junit4.createComposeRule
 import androidx.test.core.app.ApplicationProvider
 import org.dhis2.lazyActivityScenarioRule
 import org.dhis2.usescases.BaseTest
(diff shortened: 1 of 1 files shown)


dhis2-mobile-program-rules/src/main/java/org/dhis2/mobileProgramRules/RuleEngineExtensions.kt
```diff
@@ -357,6 +358,7 @@ fun ProgramRuleAction.toRuleEngineObject(): RuleAction {
 }
 
 fun ProgramRuleVariable.toRuleVariable(
+    d2: D2,
     attributeRepository: TrackedEntityAttributeCollectionRepository,
     dataElementRepository: DataElementCollectionRepository,
```
andresmr: You are adding D2 to use it in `fetchOptions` but you already have `attributeRepository` and `dataElementRepository` why not add `optionCollectionRepository` instead the full D2 and you already have the three collections?:
Something like:

Image

What do you think?

andresmr (changes requested):
Can you split the pull request into two. Leaving the tests removal in the second one. In order to pass the PR size

andresmr (approved):

PR #4535, opened 2025-12-09
fix: [ANDROAPP-7211] Collection contains no element matching the predicate

Balcan:
## Description


app/src/main/java/org/dhis2/usescases/teiDashboard/dashboardfragments/teidata/TEIDataFragment.kt
```diff
@@ -654,7 +654,7 @@ class TEIDataFragment :
             val intent = Intent(activity, EventInitialActivity::class.java)
             val bundle = Bundle()
 
-            bundle.putString(Constants.PROGRAM_UID, model.currentProgram().uid())
+            bundle.putString(Constants.PROGRAM_UID, model.currentProgram()?.uid())
```
andresmr: what could happen if this uid is null?. Could it provoke any error?. If so could we implement a way to inform the user at some point?

andresmr (approved):

mmmateos (approved):

PR #4602, opened 2026-01-20
task: [ANDROAPP-7449] Add InputCustomIntent to TrackerInputProvider

Balcan:
## Description


form/src/main/java/org/dhis2/form/ui/provider/inputfield/CustomIntentProvider.kt
```diff
@@ -80,15 +79,16 @@ fun ProvideCustomIntentInput(
                         FormIntent.OnSave(
                             it.fieldUid,
                             it.value,
-                            fieldUiModel.valueType,
+                            null,
```
xavimolloy: what is the reason behind not sending the valueType here?

PR #4638, opened 2026-02-06
feat: [ANDROAPP-7459][1/6] Added new worker to update sync settings [skip size]

Balcan:
## Description


sync/src/commonMain/kotlin/org/dhis2/mobile/sync/domain/SyncSettings.kt
```diff
@@ -0,0 +1,31 @@
...
+                    currentMetadataSyncPeriod !is SyncPeriod.Manual
+
+            if (metadataPeriodChangedFromManual) {
+                syncBackgroundJobAction.launchMetadataSync(currentMetadataSyncPeriod.toSeconds())
+                syncBackgroundJobAction.cancelSyncSettings()
+            }
+
+            Result.success(Unit)
+        } catch (e: Exception) {
+            Result.failure(e)
```
andresmr: What do you think if we try to return always `DomainError` instead of Exception? in this way we can Tailor better the presentation response. Check `PinViewModel` as an example.

mmmateos (approved):

PR #4648, opened 2026-02-11
test: [ANDROAPP-7459] [5/6] Added tests [skip size]

Balcan:
## Description


app/src/main/java/org/dhis2/data/service/SyncMetadataWorker.java
```diff
@@ -75,10 +74,7 @@ public Result doWork() {
 
             long init = System.currentTimeMillis();
             try {
```
mmmateos: Shouldn't this try catch be removed?

PR #4733, opened 2026-03-16
build: [ANDROAPP-7455] Improve Program Rule performance in forms [skip size]

Balcan:
## Description


form/src/main/java/org/dhis2/form/ui/provider/inputfield/AgeProvider.kt
```diff
@@ -96,11 +95,11 @@ fun ProvideInputAge(
                         title = fieldUiModel.label,
                         inputStyle = inputStyle,
                         isRequired = fieldUiModel.mandatory,
-                        dateOfBirthLabel = resources.getString(R.string.date_birth),
-                        orLabel = resources.getString(R.string.or),
-                        ageLabel = resources.getString(R.string.age),
-                        cancelText = resources.getString(R.string.cancel),
-                        acceptText = resources.getString(R.string.ok),
+                        dateOfBirthLabel = stringResource(R.string.date_birth),
```
andresmr: could this make test fail (if there are) or avoid testing in the future? given that we are not going to be able to mock the access to resources

andresmr (changes requested):
please check unit tests in the CI

form/src/main/java/org/dhis2/form/ui/provider/inputfield/AgeProvider.kt
Balcan: It does not affect the tests. It is the recommended way of accessing string resources from composables. We have other composables this way and the tests are fine :)

andresmr (approved):

PR #4762, opened 2026-03-25
refactor: [ANDROAPP-7527] change granular sync's state from RxJava to coroutines [skip size]

ferdyrod:
## Description

### DIFF
--- a/app/src/main/java/org/dhis2/utils/granularsync/GranularSyncRepository.kt
+++ b/app/src/main/java/org/dhis2/utils/granularsync/GranularSyncRepository.kt
@@ -1,12 +1,14 @@
 package org.dhis2.utils.granularsync
 
-import io.reactivex.Single
+import kotlinx.coroutines.async
+import kotlinx.coroutines.coroutineScope
 import kotlinx.coroutines.withContext
 import org.dhis2.R
 import org.dhis2.commons.bindings.categoryOptionCombo
 import org.dhis2.commons.bindings.countEventImportConflicts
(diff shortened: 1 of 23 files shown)


app/src/main/java/org/dhis2/utils/granularsync/GranularSyncRepository.kt
```diff
@@ -61,106 +60,119 @@ class GranularSyncRepository(
...
-            .zip(
-                getState(),
-                getLastSynced(),
-            ) { state, lastSync ->
-                buildUiState(
-                    forcedState.takeIf { it != null } ?: state,
-                    lastSync,
-                )
-            }.blockingGet()
+    suspend fun getUiState(forcedState: State? = null): SyncUiState =
```
andresmr: do you think it is a good practice to show the uistate from the repository?, could this be a domain syncState?

app/src/main/java/org/dhis2/utils/granularsync/MissingSyncTargetException.kt
```diff
@@ -0,0 +1,5 @@
+package org.dhis2.utils.granularsync
+
+class MissingSyncTargetException(
+    val uiState: SyncUiState,
```
andresmr: Custom Exceptions look to me more a business object, domain shouldn't know ui

andresmr (changes requested):

andresmr (approved):

PR #4783, opened 2026-03-31
fix: [ANDROAPP-7550] Update filters in main thread

xavimolloy:
## Description

Link the [JIRA issue](https://dhis2.atlassian.net/browse/ANDROAPP-7550).

### DIFF
--- a/app/src/main/java/org/dhis2/usescases/settings/DeleteUserData.kt
+++ b/app/src/main/java/org/dhis2/usescases/settings/DeleteUserData.kt
@@ -7,12 +7,13 @@ import org.dhis2.commons.viewmodel.DispatcherProvider
 import org.dhis2.data.service.workManager.WorkManagerController
 import java.io.File
 
-class DeleteUserData(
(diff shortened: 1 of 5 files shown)


app/src/main/java/org/dhis2/usescases/settings/DeleteUserData.kt
```diff
@@ -10,8 +12,10 @@ class DeleteUserData(
     private val filterManager: FilterManager,
     private val preferencesProvider: PreferenceProvider,
 ) {
-    fun wipeCacheAndPreferences(file: File?) {
-        filterManager.clearAllFilters()
+     suspend fun wipeCacheAndPreferences(file: File?, dispatcherProvider: DispatcherProvider) {
```
andresmr: what do you think about injecting the dispatcher in `DeleteUserData` class instead of the method?

andresmr (changes requested):

andresmr (approved):

PR #4788, opened 2026-04-08
chore: [ANDROAPP-7540] Adds the compose multiplatform dependencies directly

ferdyrod:
## Description


andresmr (commented):
Can you explain a bit how adding the dependencies solve the issue? what what was the problem then?

Balcan (approved):

ferdyrod:
> Can you explain a bit how adding the dependencies solve the issue? what what was the problem then?

Since jetbrains deprecated the dependencies accessors from the compose multiplatform plugin, we now need to add them directly.

PR #4795, opened 2026-04-13
fix: [ANDROAPP-6516] update event when synced

mmmateos:
## Description

Link the [JIRA issue](https://dhis2.atlassian.net/browse/ANDROAPP-<ANDROAPP-6516>).

### DIFF
--- a/app/src/main/java/org/dhis2/data/service/SyncRepositoryImpl.kt
+++ b/app/src/main/java/org/dhis2/data/service/SyncRepositoryImpl.kt
@@ -1,8 +1,6 @@
 package org.dhis2.data.service
 
-import io.reactivex.Observable
 import org.hisp.dhis.android.core.D2
-import org.hisp.dhis.android.core.arch.call.D2Progress
 import org.hisp.dhis.android.core.common.State
 import org.hisp.dhis.android.core.event.Event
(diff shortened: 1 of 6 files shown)


app/src/main/java/org/dhis2/data/service/SyncRepositoryImpl.kt
```diff
@@ -57,13 +55,16 @@ class SyncRepositoryImpl(
...
             .upload()
 
-    override fun downLoadEvent(eventUid: String): Observable<out D2Progress> =
-        d2
-            .eventModule()
-            .eventDownloader()
-            .byUid()
-            .eq(eventUid)
-            .download()
+    override fun downLoadEvent(
```
andresmr, line 58: who's using this method? shouldn't be a `suspend` in order to call it from async?

mmmateos: It's used with a Completable from SyncPresenter in SyncGranularWorker. I can check if possible to refactor it with coroutines

andresmr: OK, yes I think it can be a good improvement

andresmr (approved):

xavimolloy (approved):

PR #4822, opened 2026-04-22
fix: [ANDROAPP-7520] bind openIdController to the Fragment everytime the Activity is created

xavimolloy:
…created (#4787)

* fix: bind openIdController to the Fragment everytime the Activity is created

* refactor: register OpenIdControllerImpl as both concrete type and interface in LoginModule

* refactor: implement unbind() in OpenIdControllerImpl.kt to clean up resources and prevent memory leaks.

* refactor: simplify OpenIdController dependency injection

* refactor: move and rename OpenID string resources

* refactor: check FragmentActivity on bind

* refactor: optimize imports

* refactor: optimize imports

* Update app/src/main/java/org/dhis2/usescases/login/LoginActivity.kt

* feat: Enabling openid for debug mode

* feat: Refactor OpenID configuration to support multiple authentication types

* fix: add default button txt when there is no text configured

---------

## Description


login/src/androidMain/kotlin/org/dhis2/mobile/login/authentication/OpenIdControllerImpl.kt
```diff
@@ -0,0 +1,99 @@
...
+import org.hisp.dhis.android.core.user.openid.IntentWithRequestCode
+import org.jetbrains.compose.resources.getString
+
+class OpenIdControllerImpl : OpenIdController {
+    private var openIDRequestCode = -1
+    private var onResultCallback: (suspend (Result<Any>) -> Unit)? = null
+    private var openIdLauncher: ActivityResultLauncher<Intent>? = null
+    private var lifecycleScope: LifecycleCoroutineScope? = null
+
+    override fun bind(context: Any) {
```
Balcan, line 24: I guess this Any is for the interface to be Multiplatform. I think there could be a better way:
- Remove the context from the bind function
- Use the context as a parameter in the constructor

This way the context is tied to the Android implementation. What do you think?

Balcan (changes requested):

Balcan (approved):

PR #4830, opened 2026-04-27
chore: [ANDROAPP-7592] adapt to SDK domain model changes

taridepaco:
In the context of the transition towards a KMP, the domain models need to be migrated from Java to Kotlin. The SDK has started this process and small changes are expected to be done, in this case, just  nullable uid types need to be changed to non-null in some tests.

Also, from another PR merge, other changes are required concerning the PI disaggregation metadata download, which added two new mandatory fields to the PI model. Some tests need to be updated by adding the new parameters.

Link the [JIRA issue](https://dhis2.atlassian.net/browse/ANDROAPP-7592).

### DIFF
--- a/app/src/main/java/org/dhis2/usescases/teiDashboard/dashboardfragments/indicators/BaseIndicatorRepository.kt
+++ b/app/src/main/java/org/dhis2/usescases/teiDashboard/dashboardfragments/indicators/BaseIndicatorRepository.kt
@@ -13,6 +13,7 @@ import org.dhis2.mobileProgramRules.RuleConstants
 import org.dhis2.mobileProgramRules.RuleEngineHelper
 import org.hisp.dhis.android.core.D2
 import org.hisp.dhis.android.core.arch.helpers.UidGeneratorImpl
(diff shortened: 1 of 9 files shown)


andresmr (approved):

app/src/main/java/org/dhis2/usescases/teiDashboard/dashboardfragments/indicators/BaseIndicatorRepository.kt
```diff
@@ -26,6 +27,9 @@ abstract class BaseIndicatorRepository(
     open val programUid: String,
     open val resourceManager: ResourceManager,
 ) : IndicatorRepository {
+    private val defaultCategoryCombo =
+        d2.categoryModule().categoryCombos().byIsDefault().eq(true).blockingGet().first()
```
Balcan: This can lead to an ANR... It would be better to have this inside the getIndicators function.

Balcan (changes requested):

Balcan (approved):

andresmr (approved):

PR #4835, opened 2026-04-28
fix: [ANDROAPP-7569] Set default sync periods if ASWA is not configured

Balcan:
## Description


xavimolloy (approved):

andresmr (approved):

vgarciabnz:
It seems the sync job is not scheduled. It can be tested using a DHIS2 instance that has no "ANDROID_SETTINGS_APP" entry in the datastore. For example, the instance https://play.dhis2.org/demo does not have this entry by default.

Steps to reproduce:
1. Use the Android app to login into an instance without the entry in the datastore (like the demo instance).
2. Go to Settings and check the sync jobs. It says they are scheduled, but there is not a value for Next sync on. 

The property "Next sync on" only appears in the debug APK, so the debug APK must be used for testing.

PR #4860, opened 2026-05-12
build: Migrate App to kotlin [skip size]

Balcan:
## Description


app/src/main/java/org/dhis2/App.kt
```diff
@@ -0,0 +1,301 @@
...
+        setUpServerComponent()
+
+        this(getD2Configuration(this))
+
+        initCrashController()
+        setUpRxPlugin()
+        initCustomCrashActivity()
+    }
+
+    fun initCrashController() {
```
andresmr, line 129: can you think on a way of refactor this to meet with SonarCloud requirements?

andresmr (changes requested):

PR #4875, opened 2026-05-21
fix: [ANDROAPP-7634] SMS module is accidentally enabled when ASWA is installed

Balcan:
## Description

### DIFF
--- a/sync/src/androidMain/kotlin/org/dhis2/mobile/sync/data/AndroidSyncRepository.kt
+++ b/sync/src/androidMain/kotlin/org/dhis2/mobile/sync/data/AndroidSyncRepository.kt
@@ -280,36 +280,56 @@ class AndroidSyncRepository(
             Result.success(Unit)
         }
 
-    override suspend fun setUpSMS(): Result<Unit> =
+    override suspend fun setUpSMS(): Result<Boolean> =
         execute {
+            var enableSMS = false
             d2.settingModule().generalSetting().blockingGet()?.let { globalSettings ->
                 globalSettings
                     .smsGateway()
-                    ?.let {
+                    ?.let { gateway ->
+                        enableSMS = true
                         d2
                             .smsModule()
                             .configCase()
-                            .setGatewayNumber(it)
+                            .setGatewayNumber(gateway)
                             .blockingAwait()
                     }
-                globalSettings.smsResultSender()?.let {
+                globalSettings.smsResultSender()?.let { smsResult ->
                     d2
                         .smsModule()
                         .configCase()
-                        .setConfirmationSenderNumber(it)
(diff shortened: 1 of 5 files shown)


sync/src/androidMain/kotlin/org/dhis2/mobile/sync/data/AndroidSyncRepository.kt
```diff
@@ -280,36 +280,56 @@ class AndroidSyncRepository(
             Result.success(Unit)
         }
 
-    override suspend fun setUpSMS(): Result<Unit> =
+    override suspend fun setUpSMS(): Result<Boolean> =
         execute {
+            var enableSMS = false
```
vgarciabnz: This default variable would disable the SMS module unless the ASWA is configured and the gatewayNumber is defined, which prevent the use of the manual activation of the SMS module in the Android app.

I am not sure how to handle it properly to keep supporting the manual activation of the SMS module. Maybe we could deactivate it only if the ASWA is defined and the getawayNumber is not defined. In this way, we would have a way to remotely deactivate the SMS module while respecting the manual activation: the manual activation would only work when the ASWA is not installed.

jason-p-pickering: My feeling on this is if the AWSA is enabled, that implies a level of central control. Without the admin specifically entering an SMS Gateway it makes little sense. Alternatively, a separate property would be needed in the AWSA to explicitly disable the SMS module (off by default). Admins who wanted to use the SMS module would then need to both toggle this setting and/or set the SMS gateway number. Perhaps there are use cases for users setting their own gateway number, I'm just not aware of them. I think the use case we should definitely support however is the hot path, namely when the SMS module is disabled for all users centrally by the admin.

vgarciabnz: It doesn't seem very likely that the users are entering their own gateway number, but it is something that is doable in the Android app (settings screen) and it should keep working.

I have the same feeling with the ASWA. If the ASWA is defined, it implies a central control from the admin. So we would have:
- if the ASWA is installed, we check the gatewayNumber. If defined, the SMS module is enabled; if not, it is disabled. This was the behavior before 3.4.0.
- if the ASWA is not installed, we do not assume anything and the SMS module is enabled/disabled only if the user does it through the UI.

Does it sound good?

jason-p-pickering: Yeah agree . Let's document it but seems to make sense to me!

xavimolloy (approved):

vgarciabnz (approved):

PR #4907, opened 2026-06-10
fix: [ANDROAPP-7661] condition race to sync image using granular sync 

xurxodev:
Link the [JIRA issue](https://dhis2.atlassian.net/browse/ANDROAPP-7661).
  
## Problem

When a user triggers a granular sync (TEI, event, or program), SyncPresenterImpl combines the data download with the file/image download using mergeWith, which subscribes to both Observables concurrently.

The file resource downloader queries the local database for TrackedEntityAttributeValue/TrackedEntityDataValue referencing pending file resources, but if that query runs before the TEI/event/program has finished being persisted, it finds nothing pending and the image download completes without downloading anything — silently, with no error and no FileResource created locally.
  
## Solution

Replace mergeWith with concatWith in syncGranularTEI, syncGranularEvent, and syncGranularProgram. With concatWith, the file download Observable is not subscribed until the data download Observable completes, guaranteeing the TEI/event/program is fully persisted before the file resource downloader queries for pending images — matching the same sequential behavior already used by the full background sync (SyncDataWorker).

### DIFF
--- a/app/src/main/java/org/dhis2/data/service/SyncPresenterImpl.kt
+++ b/app/src/main/java/org/dhis2/data/service/SyncPresenterImpl.kt
@@ -60,7 +60,7 @@ class SyncPresenterImpl(
         return syncRepository
             .downLoadEvent(eventUid)
             .map { it as D2Progress }
-            .mergeWith(syncRepository.downloadEventFiles(eventUid))
+            .concatWith(syncRepository.downloadEventFiles(eventUid))
     }
 
     override fun blockSyncGranularProgram(programUid: String): ListenableWorker.Result {
@@ -180,7 +180,7 @@ class SyncPresenterImpl(
                 syncRepository.downloadEventProgram(uid)
             }
         }?.map { it as D2Progress }
-            ?.mergeWith(syncRepository.downloadProgramFiles(uid))
+            ?.concatWith(syncRepository.downloadProgramFiles(uid))
             ?: Observable.empty()
 
     override fun syncGranularTEI(uid: String): Observable<D2Progress> {
@@ -194,9 +194,7 @@ class SyncPresenterImpl(
         return syncRepository
             .downloadTei(teiUid, programUid)
             .map { it as D2Progress }
-            .mergeWith(
-                syncRepository.downloadTeiFiles(teiUid, programUid),
-            )
+            .concatWith(syncRepository.downloadTeiFiles(teiUid, programUid))
     }
 
     override fun syncGranularDataSet(uid: String): Observable<D2Progress> =


app/src/main/java/org/dhis2/data/service/SyncPresenterImpl.kt
```diff
@@ -60,7 +60,7 @@ class SyncPresenterImpl(
         return syncRepository
             .downLoadEvent(eventUid)
             .map { it as D2Progress }
-            .mergeWith(syncRepository.downloadEventFiles(eventUid))
+            .concatWith(syncRepository.downloadEventFiles(eventUid))
```
ferdyrod, line 63: Correct fix. One minor trade-off worth noting: `concatWith` means file downloads now happen strictly after the data download completes, increasing total wall-clock time for a granular event sync. Correctness matters more than latency here, but worth documenting so future readers do not accidentally revert this to `mergeWith`.

app/src/main/java/org/dhis2/data/service/SyncPresenterImpl.kt
```diff
@@ -194,9 +194,7 @@ class SyncPresenterImpl(
         return syncRepository
             .downloadTei(teiUid, programUid)
             .map { it as D2Progress }
-            .mergeWith(
-                syncRepository.downloadTeiFiles(teiUid, programUid),
-            )
+            .concatWith(syncRepository.downloadTeiFiles(teiUid, programUid))
```
ferdyrod, line 197: No test added for this fix. A unit test for `syncGranularTEI` (and its event/program siblings) could assert **ordering** — that `downloadTei` completes before `downloadTeiFiles` is subscribed. With RxJava's `TestObserver` and a `TestScheduler` you can verify subscription order without actually racing. Without a test, this is a silent regression risk if someone later changes the observable chain.

app/src/main/java/org/dhis2/data/service/SyncPresenterImpl.kt
```diff
@@ -194,9 +194,7 @@ class SyncPresenterImpl(
...
     }
 
     override fun syncGranularDataSet(uid: String): Observable<D2Progress> =
```
ferdyrod, line 200: `syncGranularDataSet` is not affected: it does not call a `downloadDataSetFiles` equivalent, so it does not share the same race pattern. No change needed here — just confirming the fix scope is complete.

ferdyrod (commented):
## Review: ANDROAPP-7661 — Race condition fix for granular sync image download

**Overall: correct and well-scoped fix.** The root cause is accurately identified and the solution is minimal. Replacing `mergeWith` with `concatWith` guarantees the data download fully completes (and is persisted to the local DB) before the file resource downloader queries for pending images — this mirrors the sequential behaviour already used in `SyncDataWorker` for full background sync.

A few observations inline.

ferdyrod:
Hi @xurxodev This was just a test review made by AI, you can disregard the comment posted. The only thing that we are now requiring is the verification of signed commits. Please configure it and update the commit with your signature. Thanks

xurxodev:
@ferdyrod ready

ferdyrod (approved):

andresmr (approved):

xavimolloy:
Hi @xurxodev in order to move the process along I have created a new [PR ](https://github.com/dhis2/dhis2-android-capture-app/pull/4918) that started from  and points towards the release branch. this should also fix the failed jobs and conflicts problem (I have named you as the original author of the fix 😄. Closing this PR now, thanks for the contribution!

xavimolloy:
Fix contained in https://github.com/dhis2/dhis2-android-capture-app/pull/4918

PR #4908, opened 2026-06-11
fix build warnings [skip size]

Balcan:
## Description


andresmr (changes requested):
Can you check the unit tests that are failing

PR #4912, opened 2026-06-12
fix: [ANDROAPP-7666] don't show "mark as complete?" dialog for completed events with no issues

xurxodev:
Link the [JIRA issue](https://dhis2.atlassian.net/browse/ANDROAPP-7666).

## Problem

When viewing or saving an already-completed event with no field issues, the app always shows "Saved! Do you want to mark this form as complete?", even though the event is already completed. This happens for any program, not just one specific case.

The cause is in `FormViewModel.kt`: the branch that handles already-completed events computes the "just finish, no dialog" result for events with no issues, but never actually returns it — it always falls through to showing the completion dialog instead.

  ## Solution

Return the "just finish" result when a completed event has no issues, matching the same pattern already used for skipped events right below it. A completed event with no issues now simply exits with no popup, while a completed event that still has warnings or errors keeps showing the dialog as before.


Balcan (approved):

xavimolloy (approved):

ferdyrod:
Hi @xurxodev, please update the branch with the latest updates from develop.

xurxodev:
@ferdyrod  Done, branch updated with the latest changes from develop

PR #4921, opened 2026-06-18
build: [ANDROAPP-7621] Move CI to github actions [skip size]

Balcan:
## Description

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -53,7 +53,7 @@ jobs:
 
   unit-tests:
     runs-on: ubuntu-latest
-    needs: setup
+    needs: lint-check
     steps:
       - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10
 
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -0,0 +1,134 @@
...
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  ci:
+    runs-on: ubuntu-latest
+    if: ${{ !contains(github.event.pull_request.title, '[skip ci]') && !contains(github.event.pull_request.body, '[skip ci]') }}
+
+    steps:
+      - uses: actions/checkout@v4
```
andresmr: is there any new version to avoid node.js 20 deprecation message in the pipeline?:

```
Complete job
Node.js 20 is deprecated. The following actions target Node.js 20 but are being forced to run on Node.js 24: actions/checkout@v4. For more information see: https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/
```

Balcan: It seems v7 is available. We can upgrade it to get rid of that message

.github/workflows/ci.yml
```diff
@@ -0,0 +1,274 @@
...
+      - name: Check PR Size
+        if: ${{ !contains(github.event.pull_request.title, '[skip size]') && !contains(github.event.pull_request.body, '[skip size]') }}
+        env:
+          GIT_BRANCH: ${{ github.head_ref || github.ref_name }}
+          GIT_BRANCH_DEST: ${{ github.base_ref || '' }}
+        run: |
+          chmod +x ./scripts/check_pr_size.sh
+          ./scripts/check_pr_size.sh
+
+      - name: Lint Check
```
andresmr, line 54: I would extrack Lint check from Setup and add an independent job like check style or so, maybe we can add more in the future in pararell

andresmr (changes requested):

andresmr (changes requested):
can you check and apply security review?

ferdyrod (approved):

xavimolloy (approved):

andresmr (approved):

PR #4957, opened 2026-07-02
feat: [ANDROAPP-7605] Add granular sync use-case for datasets and data values

Balcan:
## Description

### DIFF
--- a/sync/src/androidMain/kotlin/org/dhis2/mobile/sync/data/SyncDataSetRepositoryImpl.kt
+++ b/sync/src/androidMain/kotlin/org/dhis2/mobile/sync/data/SyncDataSetRepositoryImpl.kt
@@ -1,0 +1,43 @@
+package org.dhis2.mobile.sync.data
+
+import org.dhis2.mobile.sync.domain.DataSetUid
+import org.hisp.dhis.android.core.D2
+
+internal class SyncDataSetRepositoryImpl(
+    private val d2: D2,
+) : SyncDataSetRepository {
+    override suspend fun uploadDataSet(dataSetUid: DataSetUid) {
+        d2
+            .dataSetModule()
+            .dataSetInstances()
+            .byDataSetUid()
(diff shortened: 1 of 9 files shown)


sync/src/androidMain/kotlin/org/dhis2/mobile/sync/data/SyncDataSetRepositoryImpl.kt
```diff
@@ -0,0 +1,43 @@
...
+    override suspend fun uploadCompleteRegistration(dataSetUid: DataSetUid) {
+        d2
+            .dataSetModule()
+            .dataSetCompleteRegistrations()
+            .byDataSetUid()
+            .eq(dataSetUid)
+            .blockingUpload()
+    }
+
+    // TODO: The sdk does not provide methods to download data values for a given dataset
```
andresmr, line 39: is there anything to do here or could this be removed?

sync/src/androidMain/kotlin/org/dhis2/mobile/sync/data/SyncDataSetRepositoryImpl.kt
```diff
@@ -0,0 +1,43 @@
...
+    override suspend fun downloadDataSet(dataSetUid: DataSetUid) {
```
andresmr, line 40: why do you create this method if is empty? is it being used?

ferdyrod (approved):

andresmr (approved):

sync/src/androidMain/kotlin/org/dhis2/mobile/sync/data/SyncDataSetRepositoryImpl.kt
Balcan: The SDK lacks these methods, but they are important. I'll be opening a ticket in the sdk to add these methods so we can use them.

PR #4966, opened 2026-07-08
feat: [ANDROAPP-7605] Use granular sync worker from sync module [skip size]

Balcan:
## Description


app/src/main/java/org/dhis2/data/service/SyncGranularWorker.kt
```diff
@@ -50,6 +49,7 @@ import javax.inject.Inject
 private const val GRANULAR_CHANNEL = "sync_granular_notification"
 private const val SYNC_GRANULAR_ID = 8071988
 
+@Deprecated("Use org.dhis2.mobile.sync.data.SyncGranularWorker")
```
andresmr: isn't it `org.dhis2.mobile.sync.data.GranularSyncWorker`?

andresmr (changes requested):

app/src/main/java/org/dhis2/data/service/SyncGranularWorker.kt
Balcan: probably haha

mmmateos (approved):

andresmr (approved):

PR #4970, opened 2026-07-10
feat: [ANDROAPP-7686] ask to add pin on oauth logout

mmmateos:
## Description

### DIFF
--- a/app/src/main/java/org/dhis2/usescases/main/MainViewModel.kt
+++ b/app/src/main/java/org/dhis2/usescases/main/MainViewModel.kt
@@ -50,6 +50,7 @@ import org.dhis2.usescases.main.domain.ScheduleNewVersionAlert
 import org.dhis2.usescases.main.domain.UpdateInitialSyncStatus
 import org.dhis2.usescases.main.domain.model.DownloadMethod
 import org.dhis2.usescases.main.domain.model.LockAction
+import org.dhis2.usescases.main.domain.model.LogoutAction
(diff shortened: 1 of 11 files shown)


ferdyrod (approved):

app/src/main/java/org/dhis2/usescases/main/MainViewModel.kt
```diff
@@ -285,8 +288,22 @@ class MainViewModel(
         launchUseCase(dispatcher.io()) {
             matomoAnalyticsController.trackEvent(HOME, CLOSE_SESSION, CLICK)
             logOutUser().fold(
-                onSuccess = { accountCount ->
-                    _homeEffects.send(HomeEffect.GoToLogin(accountCount, false))
+                onSuccess = { logoutAction ->
+                    when (logoutAction) {
+                        LogoutAction.CreatePin -> {
+                            _homeEffects.send(HomeEffect.ShowPinDialog)
+                            logoutAfterPinSet = true
```
andresmr, line 295: what would happen if once the logoout dialog the user dismiss it and don't enter the pin? In the feature it should remain in the home cancelling the logout process. Is this covered?

mmmateos: Yes, it is covered. Here it call to display the pin dialog if proceed and the pin dialog do nothing else if is cancelled. The logout is only called on pin set

andresmr (approved):

PR #4982, opened 2026-07-21
feat: [ANDROAPP-7626] Use custom label for Organisations unit

ferdyrod:
## Description

Link the [JIRA issue](https://dhis2.atlassian.net/browse/ANDROAPP-7626).

### DIFF
--- a/app/src/main/java/org/dhis2/data/server/ServerModule.kt
+++ b/app/src/main/java/org/dhis2/data/server/ServerModule.kt
@@ -33,6 +33,8 @@ import org.dhis2.metadata.usecases.ProgramConfiguration
 import org.dhis2.metadata.usecases.TrackedEntityTypeConfiguration
 import org.dhis2.mobile.commons.files.FileController
 import org.dhis2.mobile.commons.files.FileControllerImpl
+import org.dhis2.mobile.commons.providers.CustomLabelProvider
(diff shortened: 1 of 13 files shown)


app/src/main/java/org/dhis2/data/server/ServerModule.kt
```diff
@@ -215,4 +217,8 @@ class ServerModule {
         d2: D2,
         resourceManager: ResourceManager,
     ): EventResourcesProvider = EventResourcesProvider(d2, resourceManager)
+
+    @Provides
+    @PerServer
+    fun provideCustomLabelProvider(d2: D2): CustomLabelProvider = CustomLabelProviderImpl(d2)
```
Balcan, line 223: I think we already have a Koin server module to avoid adding new dagger dependencies.

Balcan (approved):

app/src/main/java/org/dhis2/data/server/ServerModule.kt
ferdyrod: This is what I was talking about in the meeting. The provider is needed in dagger, because the FilterRepository its injected in 3 java presenters. So we would need to introduce a koin scope for those classes to use koin injection.

Personally, i don't think is worth it until those presenter are migrated to kotlin

andresmr (approved):

PR #4986, opened 2026-07-23
fix(translations): remove KMP-incompatible escaped question marks (develop)

andresmr:
Standalone cleanup: unescape `\?` → `?` in the 4 `commonskmm` Compose resource files (16 occurrences, cs/es/ru/vi). Transifex escapes `?` as `\?` for these ANDROID-typed KMP resources, which the Compose resource parser can't handle. Rendered text is unchanged. Gives a clean escaped-character baseline for the translation-sync tooling. (TEST RUN — draft.)


andresmr:
Closing — the escaped-`?` cleanup is done on main (#4987); it will reach develop via the regular main→develop merge, so a separate develop PR is redundant.

PR #4995, opened 2026-07-28
fix: [ANDROAPP-7679] Modify filter actions to check lifecycle in main activity

xavimolloy:
## Description

Link the [JIRA issue](https://dhis2.atlassian.net/browse/ANDROAPP-7679).

### DIFF
--- a/app/src/main/java/org/dhis2/usescases/main/MainActivity.kt
+++ b/app/src/main/java/org/dhis2/usescases/main/MainActivity.kt
@@ -24,6 +24,7 @@ import androidx.core.app.NotificationCompat
 import androidx.core.net.toUri
 import androidx.core.view.GravityCompat
 import androidx.drawerlayout.widget.DrawerLayout
+import androidx.lifecycle.Lifecycle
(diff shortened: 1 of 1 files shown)


Balcan (approved):

app/src/main/java/org/dhis2/usescases/main/MainActivity.kt
```diff
@@ -325,6 +326,7 @@ class MainActivity : ActivityGlobalAbstract() {
     }
 
     private fun showPeriodRequest(periodRequest: FilterManager.PeriodRequest) {
+        if (!lifecycle.currentState.isAtLeast(Lifecycle.State.RESUMED)) return
```
andresmr, line 329: is this preventing from showing the period dialog?

xavimolloy: It is not preventing, it is actually crashing at times, there  is a race condition because the listener of these requests is not context aware after the refactor, so it is calling these methods ocassionally, even when you are selecting the filters from different acitivities.

andresmr (approved):

PR #5003, opened 2026-07-31
feat: [ANDROAPP-7719] Do not close app after pin is set in home screen

Balcan:
## Description

### DIFF
--- a/app/src/main/java/org/dhis2/usescases/main/MainViewModel.kt
+++ b/app/src/main/java/org/dhis2/usescases/main/MainViewModel.kt
@@ -252,6 +252,7 @@ class MainViewModel(
                             )
                         }
                     }
+
                     SyncStatus.Succeed -> {
                         launchUseCase(dispatcher.io()) {
                             val navigationBarItems =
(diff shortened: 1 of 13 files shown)


ferdyrod (approved):

app/src/main/java/org/dhis2/usescases/main/MainViewModel.kt
```diff
@@ -516,7 +518,7 @@ class MainViewModel(
             if (logoutAfterPinSet) {
                 logOut()
             } else {
-                _homeEffects.send(HomeEffect.BlockSession)
```
andresmr: is this HomeEffect.BlockSession necessari? I think there are no more cases that uses it, could it be removed?

andresmr (changes requested):

app/src/main/java/org/dhis2/usescases/main/MainViewModel.kt
Balcan: BlockSession is used by the MainViewModel on BackPressed method.

andresmr (approved):

PR #5007, opened 2026-08-04
fix: unbreak dhis2PlayServices/dhis2Training compilation and cover them in CI [skip size]

andresmr:
## Problem

`develop` currently cannot compile two of the three app flavors:

```
* What went wrong:
Execution failed for task ':app:compileDhis2TrainingReleaseKotlin'
Execution failed for task ':app:compileDhis2PlayServicesReleaseKotlin'
   > Compilation error. See log for more details
```

ANDROAPP-7605 (#4966) removed `WorkManagerController` from `GranularSyncViewModelFactory` and
`GranularSyncPresenter`, but only the `dhis2` flavor's copy of `GranularSyncModule` was updated.
The `dhis2PlayServices` and `dhis2Training` copies still pass it, so they call the factory with 9
arguments against an 8-parameter constructor:

```
e: GranularSyncModule.kt:69 Argument type mismatch: actual type is 'WorkManagerController', but 'SMSSyncProvider' was expected.
e: GranularSyncModule.kt:70 Argument type mismatch: actual type is 'SMSSyncProvider', but 'SyncUiStateMapper' was expected.
e: GranularSyncModule.kt:71 Too many arguments for 'constructor(d2, view, repository, schedulerProvider, dispatcher, syncContext, smsSyncProvider, mapper)'.
```

`ci.yml` only ever built `:app:assembleDhis2Debug` and ran `testDhis2DebugUnitTest`, so these two
app variants were never compiled on a PR. The break only surfaced in the release-assembling
workflows — it was reported by `Verify Dependency Metadata` on #5005, which is unrelated to the
cause.

This is also the **second** time these duplicated flavor files have drifted: see #4841,
*"add imports and missing file for training build variant"*.

## Changes

- **Unbreak `dhis2PlayServices`** — drop the stale `workManagerController` param/argument, and
  reuse the module's own `provideDispatchers()` instead of a duplicated inline
  `DispatcherProvider` object.
- **Stop `dhis2Training` owning duplicate sources** — its copies of `GranularSyncModule.kt` and
  `DownloadNewVersion.kt` were byte-identical to `dhis2`'s, so the flavor now shares
  `src/dhis2/java` and both copies are deleted. Training keeps its own `AndroidManifest.xml` and
  `res/` (launcher icons, app name); only Kotlin is shared.
- **Guard it in CI** — compile the `dhis2PlayServices` and `dhis2Training` flavors in the existing
  `build-test-apks` job. Dagger runs through KSP and `compileKotlin` depends on the KSP task, so
  compiling catches both source errors and an unresolvable DI graph, without the
  resource-merge/dex/package cost of a full assemble, no extra runner and no extra Gradle cache
  warm-up.

## Notes for reviewers

- `java.setSrcDirs` / `java.directories` does **not** feed the Kotlin compilation in AGP 9 — the
  first attempt with it left Training's kapt stubs missing `GranularSyncModule`. `kotlin.srcDir`
  is the working knob, and it must be configured *after* `productFlavors`, which is what creates
  the per-flavor source sets.
- Any file added to `src/dhis2/java` from now on also compiles into `dhis2Training`. That is the
  intent — both are the non-Play build — but it is a behaviour change worth knowing.
- If you have an existing `app/build` from before this change and hit a one-off
  `kapt<Flavor>Kotlin` failure, re-run the task; stale kapt stubs from the old source-set layout
  clear themselves on the second run.

## Verification

Run locally, all passing, with `:app:clean` first so the results are not incremental artifacts:

```bash
# the three release compiles (the two that fail on develop, plus the one that passes)
./gradlew :app:compileDhis2ReleaseKotlin \
          :app:compileDhis2PlayServicesReleaseKotlin \
          :app:compileDhis2TrainingReleaseKotlin

# kapt + KSP + Dagger graph actually link for the previously-broken variants
./gradlew :app:assembleDhis2PlayServicesDebug :app:assembleDhis2TrainingDebug

# the modified CI step, exactly as written
./gradlew :app:assembleDhis2Debug :app:assembleDhis2DebugAndroidTest :form:assembleAndroidTest \
          :app:compileDhis2PlayServicesDebugKotlin :app:compileDhis2TrainingDebugKotlin --no-daemon

./gradlew ktlintCheck
```

The CI guard was also negative-tested: reintroducing the extra constructor argument makes
`:app:compileDhis2PlayServicesDebugKotlin` fail with the same three errors quoted above, so the
new step genuinely catches this class of break rather than merely passing.

Once this merges, #5005 needs `develop` merged in to turn its `Verify Dependency Metadata` check
green.

## Follow-up (not in this PR)

`dhis2PlayServices`'s `GranularSyncModule.kt` is still a near-copy of `dhis2`'s, differing only in
the `smsSyncProvider()` body. Moving the module to `src/main` and leaving just a small per-flavor
`@Provides` for `SMSSyncProvider` would remove the last of this duplication.


Balcan (approved):

.github/workflows/ci.yml
```diff
@@ -98,7 +98,13 @@ jobs:
           cache: 'gradle'
 
       - name: Build Test APKs
-        run: ./gradlew :app:assembleDhis2Debug :app:assembleDhis2DebugAndroidTest :form:assembleAndroidTest --no-daemon
+        run: |
+          ./gradlew :app:assembleDhis2Debug \
+            :app:assembleDhis2DebugAndroidTest \
+            :form:assembleAndroidTest \
+            :app:compileDhis2PlayServicesDebugKotlin \
```
andresmr, line 105: adding compile playServices and training to avoid breaking builds in the future. The cost is minimum

xavimolloy (approved):

PR #5024, opened 2026-08-07
chore: remove Glide and migrate remaining image loading to Coil

andresmr:
## Why

This PR's dependabot group bumps Glide `5.0.7` → `5.0.9`, and Glide 5.0.9 declares `minCompileSdk=37`, which is what broke `:app:checkDhis2DebugAarMetadata` on `unit-tests`, `verify-dependency-metadata` and `deployment_job`. There is no intermediate version to land on — **5.0.8 was never published** to Maven Central.

Looking at what Glide was actually doing here made removal the more attractive answer: **6 call sites, all View-based, none loading over the network.** Every load was a local file, an in-memory `Bitmap`, or a drawable resource id. Meanwhile **Coil 3 is already a dependency** — `commonskmm` exposes `coil-compose` via `api`, with `coil` and `coil-core` arriving transitively — and was being used only for `coil3.PlatformContext` as a KMP `Context` abstraction, with zero image loading. So this removes a library without adding one.

> [!IMPORTANT]
> Glide was the **only** dependency in the whole graph requiring `minCompileSdk > 36` (verified by scanning every cached AAR). With it gone, the `compileSdk 37` bump in this PR is no longer *required* — keep it if you want to be current, but it's now a choice rather than a forced move.

## What changed

**Dead and inert code removed**
- `ItemPicProvider.kt` deleted — `setItemPic` had **no callers anywhere** in the repo (2 of the 6 call sites)
- `MyAppGlideModule` deleted — an empty `AppGlideModule`. The KSP processor was registered in `:commons` while the module lived in `:app`, and `ksp` isn't transitive, so nothing was ever generated and Glide logged *"Failed to find GeneratedAppGlideModule"* at runtime

**Straight replacements**
- `QrActivity` / `QrFragment` load an already-decoded in-memory `Bitmap` — now `ImageView.setImageBitmap`, no image library involved
- `SearchTeiModelExtensions` and `TEIDataFragment` load the TEI avatar via `ImageView.load` with `CircleCropTransformation` + crossfade. Coil cancels any in-flight request for a recycled `ImageView`, matching Glide, which matters since the search one is a RecyclerView row. Both now pass `File(path)` rather than a raw path `String`, which Coil resolves unambiguously.
- `TeiMapManager` replaces `asBitmap` + `CustomTarget` with `ImageLoader.execute` in a coroutine, then `toBitmap` at the same 30dp size before handing the bitmap to `TeiMarkers.getMarker`

**Two behaviour fixes** fell out of the `TeiMapManager` rewrite, called out so they aren't mistaken for refactor noise:
1. The old code called `setSource()` from the callback of whichever load finished last **by index**, not last in time — so it commonly fired before slower loads had populated `teiImages`.
2. If the last-indexed load *failed*, `onResourceReady` never fired for it and `setSource()` was **never called at all**.

Loads now run concurrently via `async`/`awaitAll`, and `setSource()` is called exactly once, on the main dispatcher, after every load has settled. Glide also cancelled work automatically through `Glide.with(context)`; to preserve that, the load job is cancelled on the previous run and on `ON_DESTROY`.

## Verification

- `./gradlew ktlintCheck` ✅
- `./gradlew :app:assembleDhis2Debug` ✅
- `./gradlew testDebugUnitTest testDhis2DebugUnitTest testAndroidHostTest` ✅
- `com.bumptech.glide` no longer resolves on any classpath ✅

## Needs manual QA

These three surfaces are visual and not unit-testable, so they want a look on a device:
- [ ] Search list TEI avatars — scroll hard to exercise view recycling
- [ ] TEI dashboard legacy card avatar
- [ ] Map TEI markers — both a `dhis2_*` metadata icon and a real TEI photo


andresmr:
Superseded by #5026 — same commit, rebranched as `ANDROAPP-7743-remove-glide` to follow the repo's ANDROAPP-<key> branch convention and linked to ANDROAPP-7743.

PR #5043, opened 2026-08-13
fix: [ANDROAPP-7732] Checks if server has oAuth enabled to verify if "login with openId" button should be displayed

ferdyrod:
## Description

Link the [JIRA issue](https://dhis2.atlassian.net/browse/ANDROAPP-7732).

### DIFF
--- a/login/src/commonMain/kotlin/org/dhis2/mobile/login/main/ui/screen/CredentialsScreen.kt
+++ b/login/src/commonMain/kotlin/org/dhis2/mobile/login/main/ui/screen/CredentialsScreen.kt
@@ -240,6 +240,7 @@ fun CredentialsScreen(
                 onCredentialsAction = { credentialsAction ->
                     handleCredentialAction(viewModel, context, credentialsAction)
                 },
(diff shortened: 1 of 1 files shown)


login/src/commonMain/kotlin/org/dhis2/mobile/login/main/ui/screen/CredentialsScreen.kt
```diff
@@ -240,6 +240,7 @@ fun CredentialsScreen(
                 onCredentialsAction = { credentialsAction ->
                     handleCredentialAction(viewModel, context, credentialsAction)
                 },
+                hasOAuthEnabled = entryMode == CredentialsEntryMode.EXISTING_OAUTH,
```
andresmr, line 243: What would happen if it is a new OAuth account and during the login flow, the user close the custom tab, would the open id button be visible

ferdyrod: No it will not display the open Id button.

andresmr (approved):

PR #5068, opened 2026-08-24
ANDROAPP-7733-ask-the-user-to-login-online-on-expired-missing-tokens [skip size]

andresmr:
## Description

Link the [JIRA issue](https://dhis2.atlassian.net/browse/ANDROAPP-7733).

Ask user to log in when the token expires


andresmr:
This PR has been split into seven smaller ones so it can actually be reviewed. It is now a draft and stays open as the umbrella for the feature; the branch remains the source of truth until the seven have landed.

Every part is the same code, moved verbatim — the 55 changed files are byte-identical to this branch, nothing added and nothing dropped. Six of the seven come in under the 400-line limit without `[skip size]`.

### The seven parts


| Part | PR | Base | Lines |
|---|---|---|---|
| 1 | #5070 — commons: error mapping + renewal signal | `develop` | 334 |
| 2 | #5071 — sync: repositories via `withDomainErrors` | #5070 | 380 |
| 3 | #5072 — sync: expired session is not an unavailable server | #5071 | 85 |
| 4 | #5073 — sync: raise the renewal signal | #5072 | 349 |
| 5 | #5074 — login: renew from the credentials screen | #5070 | 360 |
| 6 | #5075 — app: dialog and routing | #5074 | 200 |
| 7 | #5076 — dev tool: force session expiry | `develop` | 403 (size check waived) |

The 55 changed files are byte-identical to #5068 — verified file by file — with one exception: the three-line `existingUsername = null` placeholder #5070 needs to compile on its own, which #5074 replaces.


**Why the split falls where it does**

- The SDK bump `-75` → `-77` is not separable from Part 1: the snapshot adds `AUTHENTICATED_USER_MISMATCH` to `D2ErrorCode` (two exhaustive `when`s stop compiling without it) and makes `existingUsername` a required parameter of `blockingHandleLogInResponse`. Part 1 therefore carries a three-line `existingUsername = null` placeholder in `LoginRepositoryImpl` — today's behaviour — which Part 5 replaces with the account being restored.
- Parts 3 and 4 are the two product bugs this PR found and fixed. They were the easiest things to miss in 2105 lines and are the two worth the closest reading.
- Part 2 is a mechanical five-file refactor with no behaviour change beyond the `RuntimeException` unwrapping, kept apart so it does not hide Parts 3 and 4.
- Part 7 is debug-only and depends on nothing else, so it goes straight onto `develop` and can be rejected without holding up the feature.

PR #5086, opened 2026-08-27
chore: [ANDROAPP-7758] review picture processing [skip size]

xavimolloy:
## Description

Link the [JIRA issue](https://dhis2.atlassian.net/browse/ANDROAPP-7758).


form/src/main/java/org/dhis2/form/data/FormValueStore.kt
```diff
@@ -560,21 +561,43 @@ class FormValueStore(
...
+     * data element mode, so the program they report is the one the form is being filled for. There is
+     * deliberately no lookup by record as a fallback: the record of this store is the tracked entity
+     * instance, and querying its enrollments would return one row per program it is enrolled in, with
+     * nothing to tell which of them this form belongs to.
+     *
+     * A null means neither repository was injected, which is the case for the stores the event
+     * capture and the tei dashboard build. Those only apply program rule effects and never store a
+     * file; were that to change, no program level image setting would be found and the Sdk would
+     * apply its default quality.
+     */
```
andresmr: Is this comment valuable to understand the context for `resolveProgramUid()`?.
Maybe if it needs an explanation could be more concise.
What do you think?

xavimolloy: completely agree, I will update the ticket

andresmr (approved):

PR #5087, opened 2026-08-27
fix: [ANDROAPP-7775] report the project's real test coverage

andresmr:
Coverage was reported for unit tests only, over roughly a fifth of the codebase. Two
independent causes, both silent — a missing coverage report is a warning, not an error,
and CI hid even that (`if-no-files-found: warn`, `continue-on-error: true`).

## Cause 1 — instrumented coverage never left BrowserStack

The UI suites run on BrowserStack's devices, so their JaCoCo execution data was written
there and nothing brought it back. `code-quality` also ran *alongside* those jobs rather
than after them, so it could not have used the data even if it had arrived.

## Cause 2 — the exclusion list removed 82.3% of the code

`jacoco/jacoco.gradle.kts` excluded **1803 of 2191** compiled classes in `app`. What the
report called the project's coverage was the coverage of the 18% that survived.

Measured per pattern on `app`'s `dhis2Debug` output:

| classes removed | pattern | what it actually removed |
| ---: | --- | --- |
| 1046 | `**/*$*` | every nested class, lambda and coroutine state machine |
| 664 | `**/*$*$*.*` | |
| 332 | `**/*Module*.*` | anything with "Module" in the name, not just DI modules |
| 197 | `**/*Activity*.*` | |
| 153 | `**/databinding/*.*` | (legitimately generated) |
| 127 | `**/ui/*.class` | including plain mappers that happen to live under `ui/` |
| 110 | `**/*Dialog*.*` | |

Among the casualties of `**/*$*` were 197 **named** nested types carrying real logic:
sealed subclasses (`ConvertTaskResult$Message`), nested enums (`OpenIdSession$LogOutReason`),
`$DefaultImpls` — the actual bodies of interface default methods — and builders.

## What changed

1. **Instrument debug builds** — `enableAndroidTestCoverage = true` on the debug build
   type of `app` and `form`, so instrumented runs emit `.ec` data. Debug only; release
   builds and the shipped APK are unaffected.
2. **Retrieve the data** — new `scripts/fetchBrowserstackCoverage.sh` reads
   `.devices[].sessions[].id` from `GET /app-automate/espresso/v2/builds/<id>` and
   downloads each session from `.../sessions/<session>/coverage`. `"coverage": true`
   added to the three build requests. The script never fails its caller — missing
   coverage should not turn a green test run red — and runs before the pass/fail exit so
   a failed suite still contributes what it covered.
3. **Merge before analysis** — the BrowserStack jobs upload what they retrieved,
   `unit-tests` also uploads raw `**/build/jacoco/*.exec`, and `code-quality` now waits
   on all of them, copies the `.ec` files into each reporting module and regenerates the
   XML before Sonar runs.
4. **Exclude only generated code** — the list now keeps R/BR/BuildConfig/Manifest, Data
   Binding, Dagger/Hilt, other annotation processors, `$WhenMappings`, `$$serializer`,
   `ComposableSingletons` and test classes, and nothing else. Android UI classes are no
   longer excluded **on purpose**: the instrumented suite exercises them, and hiding them
   understated the very tests that cover them.
5. **Unpin JaCoCo** — required, not incidental. See below.
6. **`verify-dependency-metadata.yml` gets a `concurrency` block** — see below.

## Required: unpinning JaCoCo

Turning on instrumentation broke the build outright:

```
Execution failed for task ':form:mergeExtDexDebugAndroidTest'
  > Failed to transform mockito-kotlin-6.1.0.jar ... useJacocoTransformInstrumentation=true
    > Execution failed for JacocoTransform:
      .../byte-buddy/1.17.7/byte-buddy-1.17.7.jar
```

`enableAndroidTestCoverage` makes AGP run `JacocoTransform` over **every** dependency on
the androidTest runtime classpath. The root build pinned all `org.jacoco` artifacts to
**0.8.10** through a `resolutionStrategy`, overriding whatever version AGP asked for, and
0.8.10 cannot instrument byte-buddy 1.17.7's bytecode. The pin predates instrumented
coverage and cannot survive it.

The pin is removed and the jacoco plugin's `toolVersion` is wired to
`libs.versions.jacoco`, so the catalog governs the report engine rather than Gradle's
bundled default. Modules that instrument (`app`, `form`) use AGP's own jacoco version for
the transform — the version AGP is tested against.

Verified locally that `:form:mergeExtDexDebugAndroidTest`, the task that failed, builds.

`gradle/verification-metadata.xml` gains the entries the newer artifacts require.

## Result — verified on CI, every job green

```
instrumented coverage files retrieved: 3
```

| job | `.ec` retrieved |
| --- | --- |
| run-ui-landscape | 1.3 MB |
| run-ui-portrait | 1.3 MB |

|  | coverage | lines measured | lines covered |
| --- | ---: | ---: | ---: |
| `develop` | 10.6% | 47,579 | 5,486 |
| **this branch** | **17.9%** | **61,224** | **12,298** |

**Read the counts, not just the percentage.** This measures **29% more code** and credits
**2.2× more covered lines** than `develop`. Roughly 6,800 more lines of genuinely tested code are credited than on `develop`, and a
large body of untested UI code that was hidden from the denominator is now measured.

Exclusions: **82.3% → 25.4%**. Classes analysed in `app`: **329 → 1074**.

## Also fixed: `Verify Dependency Metadata` pushed empty commits

That workflow had no `concurrency` block, so its runs stacked up. Each compares its
regenerated metadata against the commit it checked out rather than the branch tip, so the
slower of two overlapping runs still saw a difference after the faster one had already
committed the same fix, and pushed a commit with a byte-identical tree:

```
08:06:56  run A starts at ea6d2506c
08:23:47  run B starts at e2d3f3469
08:41:02  run A pushes 397a0a2da   (+18 lines -- legitimate)
08:46:10  run B pushes 04caca0c7   (identical tree -- empty)
```

Every such push cancelled the in-flight CI run. Four runs died this way while this work
was being verified. Serialising per ref means only the newest run survives, and it always
regenerates from the newest tip.

## Note on the form suite

ANDROAPP-7748 (#5034) retired the form instrumentation CI job while this PR was open, and
this branch adopts that fully: the BrowserStack form script is deleted, `form` is out of
the coverage staging list, and `form/build.gradle.kts` is back to develop's version —
`develop` has no `form/src/androidTest` sources left, so instrumenting that module would
cost build time for data that cannot exist.

Instrumented coverage therefore comes from two device jobs rather than three. That is why
this figure moved from 19.1% when first measured to 17.9%: the denominator is unchanged at
61,224 — only device-sourced covered lines fell. The form module's replacement JVM tests
are counted through the normal unit-test path.

## Trade-off

`code-quality` now waits on the device suites, so the Sonar result arrives later than it
used to. That is inherent to including their coverage.

## Not in this PR

- Making the metadata check compare against `origin/<branch>` rather than its own
  checkout. The `concurrency` block stops runs from overlapping, but the comparison is
  still against the wrong baseline if anything else pushes mid-run.
- Deleting the stale `SonarCloud` required status check from the rulesets — nothing
  reports it, so it can never pass. Needs an admin. (`SonarCloud Code Analysis` and
  `code-quality` are both legitimate and should stay: one is the scan, the other the
  Quality Gate verdict.)


andresmr:
Superseded by #5154 — same commits, rebased onto develop and moved to a branch matching the `ANDROAPP-<ticket>-<description>` convention. GitHub's branch-rename API needs org admin for ruleset-protected branches, so this had to be a new PR.

PR #5093, opened 2026-08-27
fix: stabilise TeiDashboardTest.shouldHandleNotesWorkflow

andresmr:
`TeiDashboardTest.shouldHandleNotesWorkflow` failed intermittently on the BrowserStack
matrix, in landscape, and passed on re-run with no code change. Two assumptions about the
soft keyboard, both landscape-specific.

## Typing

`typeNote` used `TypeTextAction`, which needs the IME. In landscape most devices open the
keyboard in fullscreen **extract mode**, which replaces the field under test with the IME's
own editor — the keystrokes never reach `noteText`.

It now uses `replaceText`, which fires the same text watchers without depending on the
keyboard at all.

## Clicking

Hiding the IME starts an animated window-inset transition. A click issued while that
animation is still running resolves the target's coordinates mid-flight and lands where the
button used to be — usually on the keyboard. `isDisplayed()` does not know a view is
obscured by the IME window, so nothing caught this.

New `BaseRobot.waitUntilViewIsStable(matcher)` polls a view's on-screen position until it
stops moving. It replaces a `waitToDebounce(300)` sleep in `clickOnClearButton` — a
hard-coded delay, which this repo forbids — and is added to `clickOnSaveButton`, which had
no protection at all.

## Verification

Passed in portrait and landscape on a full 41-test run and has not failed since. These
tests run only on BrowserStack via CI and cannot be executed locally;
`:app:assembleDhis2DebugAndroidTest` builds and ktlint passes.

### DIFF
--- a/app/src/androidTest/java/org/dhis2/common/BaseRobot.kt
+++ b/app/src/androidTest/java/org/dhis2/common/BaseRobot.kt
@@ -4,6 +4,7 @@ import android.app.Activity
 import android.app.ActivityManager
 import android.content.Context
 import android.content.Context.ACTIVITY_SERVICE
+import android.os.SystemClock
 import android.view.View
 import android.view.inputmethod.InputMethodManager
 import androidx.compose.ui.semantics.SemanticsNode
@@ -20,6 +21,7 @@ import androidx.test.espresso.UiController
 import androidx.test.espresso.ViewAction
 import androidx.test.espresso.ViewInteraction
 import androidx.test.espresso.base.DefaultFailureHandler
+import androidx.test.espresso.matcher.ViewMatchers.isDisplayed
 import androidx.test.espresso.matcher.ViewMatchers.isRoot
 import androidx.test.espresso.util.TreeIterables
 import androidx.test.platform.app.InstrumentationRegistry.getInstrumentation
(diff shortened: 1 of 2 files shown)


app/src/androidTest/java/org/dhis2/common/BaseRobot.kt
```diff
@@ -78,6 +80,45 @@ open class BaseRobot {
...
 
+    /**
+     * Waits until the view matched by [viewMatcher] has stopped moving on screen.
+     *
+     * Hiding the IME starts an animated window-inset transition. A click issued while it is
+     * still running resolves the target's coordinates mid-animation, so the tap lands
+     * wherever the button used to be - most often on the keyboard itself. This is especially
+     * easy to hit in landscape, where the IME covers most of the window. Polling the view's
+     * position waits for the actual condition instead of guessing a duration.
+     */
```
andresmr, line 88: can you simply this comment just explaining what it does?

app/src/androidTest/java/org/dhis2/usescases/datasets/DataSetDetailRobot.kt
```diff
@@ -168,4 +169,28 @@ internal class DataSetDetailRobot(
...
     }
+
+    /**
+     * Opens the completed data set instance, identified by its completion icon.
+     *
+     * Do not reach for it by index. The list is sorted by period, so the position of the
+     * completed instance depends on which periods the test created. The row also only gains
+     * its completion icon once the list has refreshed after returning from the form, so
+     * matching on the icon waits for that refresh as well.
+     */
```
andresmr: same here, don't explain was it was wrong, just how it works now

app/src/androidTest/java/org/dhis2/common/BaseRobot.kt
andresmr: Cut to what it does plus when to use it: "Waits until the view matched by [viewMatcher] has stopped moving on screen. Use before clicking a view that an animation may still be repositioning, such as one next to a keyboard that is opening or closing."

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

app/src/androidTest/java/org/dhis2/usescases/datasets/DataSetDetailRobot.kt
andresmr: Now just "Waits for the data set instance marked as completed and opens it." — the rationale about indexes and list refresh is gone.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

xavimolloy (approved):

PR #5104, opened 2026-08-31
fix: [ANDROAPP-7772] Show original file name in file resource fields [skip size]

taridepaco:
## Description

Link the [JIRA issue](https://dhis2.atlassian.net/browse/ANDROAPP-7772).

A `FILE_RESOURCE` field showed the file resource uid with its extension (`afl3jai2i4u.pdf`) instead of the original file name (`report.pdf`).

This is not an SDK bug: the SDK keeps the original name in `FileResource.name` from end to end, and deliberately renames only the file **on disk** to `<uid>.<ext>`. The app was reading the basename of `FileResource.path()` and never reading `name()`.

### Solution

A new `fileName` property on `FieldUiModel`, populated alongside `displayName`, which keeps holding the path. `displayName` is consumed as a path by `FormView.openFile`, the file weight, the image preview and the signature field, so its semantics are left untouched — if a recomputation site were ever missed, the field simply falls back to the current behaviour instead of breaking downloads or previews.

- `D2.fileResourceNameOf(uidOrPath)` resolves the name by uid and, failing that, by path. Both are needed: a form field holds the uid right after picking a file, and the path once the form is composed again — the latter is the case that was broken.
- Routed through `DisplayNameProvider`, the only collaborator already present at every recomputation site, so no constructor signature or DI wiring changes.
- Event and relationship cards had the same root cause and were leaking the absolute path to the user (`ProgramEventMapper.getData()`, `RelationshipsRepository`). They now show the name too.

### Testing

Unit tests cover name resolution by uid, by path, a missing file resource and a null name, plus the recomputation sites and both card mappers. A new instrumented test covers the input field showing the original name, falling back to `<uid>.<ext>` when there is none, and the empty state.

Manually verified on device: picking a file, reopening the form, files downloaded from the server, enrollment forms, clearing the value, opening/downloading the file, rotation, image and signature fields, aggregate data set file fields, and both card types.

### Known gap

`FormView.downloadFile` still copies the file to Downloads under its `<uid>.<ext>` name. Out of scope here; `fileResourceNameOf()` is available if we want to fix that separately.

### DIFF
--- a/form/src/main/java/org/dhis2/form/ui/provider/inputfield/InputFileProvider.kt
+++ b/form/src/main/java/org/dhis2/form/ui/provider/inputfield/InputFileProvider.kt
@@ -46,7 +46,7 @@ internal fun ProvideInputFileResource(
         supportingText = fieldUiModel.supportingText(),
         buttonText = stringResource(R.string.add_file),
         uploadFileState = uploadState,
-        fileName = file?.name,
+        fileName = fieldUiModel.fileName ?: file?.name,
         fileWeight = file?.length()?.let { fileSizeLabel(it) },
         onSelectFile = {
             uploadState = getFileUploadState(fieldUiModel.displayName, true)
--- a/app/src/main/java/org/dhis2/usescases/programEventDetail/ProgramEventMapper.kt
+++ b/app/src/main/java/org/dhis2/usescases/programEventDetail/ProgramEventMapper.kt
@@ -1,5 +1,6 @@
 package org.dhis2.usescases.programEventDetail
 
+import org.dhis2.bindings.fileResourceNameOf
 import org.dhis2.bindings.userFriendlyValue
 import org.dhis2.commons.data.EventModel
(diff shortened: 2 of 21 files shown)


form/src/main/java/org/dhis2/form/ui/provider/inputfield/InputFileProvider.kt
```diff
@@ -46,7 +46,7 @@ internal fun ProvideInputFileResource(
         supportingText = fieldUiModel.supportingText(),
         buttonText = stringResource(R.string.add_file),
         uploadFileState = uploadState,
-        fileName = file?.name,
+        fileName = fieldUiModel.fileName ?: file?.name,
```
andresmr, line 49: as we are using a `val file` variable on line 40, should be better to resolve this in the file property declaration? something like `val file = fieldUiModel.fileName ?:  fieldUiModel.displayName?.let { File(it) }?.name`

andresmr (approved):

PR #5112, opened 2026-09-02
feat: [ANDROAPP-7660] Display error messages in search results

Balcan:
## Description

### DIFF
--- a/app/src/main/java/org/dhis2/usescases/searchTrackEntity/SearchTEIViewModel.kt
+++ b/app/src/main/java/org/dhis2/usescases/searchTrackEntity/SearchTEIViewModel.kt
@@ -45,6 +45,7 @@ import org.dhis2.commons.extensions.toFriendlyDate
 import org.dhis2.commons.extensions.toFriendlyDateTime
 import org.dhis2.commons.extensions.toPercentage
 import org.dhis2.commons.filters.FilterManager
+import org.dhis2.commons.filters.sorting.SortingItem
 import org.dhis2.commons.network.NetworkUtils
 import org.dhis2.commons.resources.ResourceManager
 import org.dhis2.commons.viewmodel.DispatcherProvider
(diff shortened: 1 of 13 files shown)


xavimolloy (approved):

app/src/main/java/org/dhis2/usescases/searchTrackEntity/SearchTEIViewModel.kt
```diff
@@ -546,6 +554,32 @@ class SearchTEIViewModel(
             searchParametersUiState.copy(searchEnabled = queryDataList.isNotEmpty())
     }
 
+    /**
+     * Maps the repository's paging stream to [SearchTeiModel]s for the RecyclerView adapter.
+     * A page-load failure ([TrackedEntitySearchPagingItem.Error]) is logged and dropped from the
+     * list rather than failing the whole page load.
+     */
+    private fun Flow<PagingData<TrackedEntitySearchPagingItem>>.toSearchTeiModelPagingData(
```
andresmr, line 562: what do you think about having this extension in it's own class?.
Maybe we can create it in tracker, but it would need to refactor SearchTeiModel to this module

tracker/src/commonMain/kotlin/org/dhis2/tracker/search/model/TrackedEntitySearchItemResult.kt
```diff
@@ -1,9 +1,24 @@
...
+import org.dhis2.mobile.commons.error.DomainError
 import org.dhis2.tracker.input.model.TrackerInputType
 import org.dhis2.tracker.relationships.model.RelationshipModel
 import kotlin.time.Instant
 
+/**
+ * A single row emitted by the search paging stream. A page can partially fail (some rows load
+ * fine, others don't), so failures are carried per item instead of failing the whole page.
+ */
+sealed interface TrackedEntitySearchPagingItem {
```
andresmr: Can you extract it to it's own file?

andresmr (changes requested):

app/src/main/java/org/dhis2/usescases/searchTrackEntity/SearchTEIViewModel.kt
Balcan: This requires too many changes and additions. I think we can leave it as it is and it will be easier once we refactor the search and we start using usecases.

tracker/src/commonMain/kotlin/org/dhis2/tracker/search/model/TrackedEntitySearchItemResult.kt
Balcan: Done!

andresmr (approved):

PR #5115, opened 2026-09-03
fix: [ANDROAPP-7675] Display not found feedback for InputImage

xavimolloy:
## Description

Link the [JIRA issue](https://dhis2.atlassian.net/browse/ANDROAPP-7675).

### DIFF
--- a/aggregates/src/androidMain/kotlin/org/dhis2/mobile/aggregates/data/DataSetInstanceRepositoryImpl.kt
+++ b/aggregates/src/androidMain/kotlin/org/dhis2/mobile/aggregates/data/DataSetInstanceRepositoryImpl.kt
@@ -1,5 +1,6 @@
 package org.dhis2.mobile.aggregates.data
 
+import org.dhis2.bindings.FILE_NOT_FOUND
 import org.dhis2.commons.bindings.dataElement
 import org.dhis2.commons.periods.data.PeriodLabelProvider
 import org.dhis2.mobile.aggregates.data.mappers.toCustomTitle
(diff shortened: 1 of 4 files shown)


aggregates/src/androidMain/kotlin/org/dhis2/mobile/aggregates/data/DataSetInstanceRepositoryImpl.kt
```diff
@@ -1308,13 +1308,13 @@ internal class DataSetInstanceRepositoryImpl(
...
 
-    override suspend fun getFilePath(fileUid: String): String? =
+    override suspend fun getFilePath(fileUid: String): String =
         d2
             .fileResourceModule()
             .fileResources()
             .uid(fileUid)
             .blockingGet()
-            ?.path()
+            ?.path() ?: "fileNotFound"
```
andresmr: could this hardcoded text be moved to strings?

commons/src/main/java/org/dhis2/commons/bindings/ValueExtensions.kt
```diff
@@ -110,7 +110,7 @@ fun checkValueTypeValue(
                 .fileResources()
                 .uid(value)
                 .blockingGet()
-                ?.path() ?: ""
+                ?.path() ?: "fileNotFound"
```
andresmr: same here, maybe you could reuse the same string resource

andresmr (changes requested):

aggregates/src/androidMain/kotlin/org/dhis2/mobile/aggregates/data/DataSetInstanceRepositoryImpl.kt
xavimolloy: This string is not displayed to the user at any point, if the code reaches that point it means that there is a value to the  attr or DE but the file was not downloaded, fileNotFound will cause the inputImage to display the not found default image, but the string will never be displayed

Balcan (approved):

andresmr (approved):

PR #5132, opened 2026-09-08
fix: [ANDROAPP-7697] User has access to one program or dataset

Balcan:
## Description


xavimolloy (approved):

app/src/main/java/org/dhis2/usescases/main/program/ProgramModule.kt
```diff
@@ -41,6 +42,7 @@ class ProgramModule(
             filterManager,
             syncStatusController,
             schedulerProvider,
+            NavigateToSingleProgram(programRepository),
```
andresmr, line 45: why is not injected like the other classes?

Balcan: It is only used in this class. I did not want to increase the dagger debt

PR #5139, opened 2026-09-09
fix: [ANDROAPP-7786] Token expired and Sync successful messages at the same time

Balcan:
## Description


xavimolloy (approved):

sync/src/commonMain/kotlin/org/dhis2/mobile/sync/model/GranularSyncAction.kt
andresmr, line 1: given that we have a sync module, what do you think if this action can be created in the sync module? would it make sense?

andresmr (changes requested):

andresmr (approved):

PR #5140, opened 2026-09-09
fix: [ANDROAPP-7785] Update sync status dialog during sync

xavimolloy:
## Description

Link the [JIRA issue](https://dhis2.atlassian.net/browse/ANDROAPP-7785).

### DIFF
--- a/app/src/main/java/org/dhis2/utils/granularsync/GranularSyncPresenter.kt
+++ b/app/src/main/java/org/dhis2/utils/granularsync/GranularSyncPresenter.kt
@@ -31,12 +31,10 @@ import androidx.lifecycle.MutableLiveData
 import androidx.lifecycle.ViewModel
 import androidx.lifecycle.viewModelScope
 import io.reactivex.disposables.CompositeDisposable
-import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.flow.MutableStateFlow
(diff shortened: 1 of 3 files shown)


Balcan (approved):

app/src/main/java/org/dhis2/utils/granularsync/GranularSyncPresenter.kt
```diff
@@ -164,13 +163,13 @@ class GranularSyncPresenter(
...
         }
-        return observeWorkInfo()
     }
 
     fun observeWorkInfo() =
         syncBackgroundJobAction.observeGranularJob(
             when (syncContext.conflictType()) {
-                ALL -> Constants.INITIAL_SYNC
+                // Matches the unique work name enqueued by launchDataSync(0) in initGranularSync
+                ALL -> DATA_SYNC_NOW
```
andresmr, line 172: if this is a defect, do we now when it was changed and why? if it wasn't change, why it stop working?

xavimolloy: The problem was not with the constant as such, but with the observer. And we lost it when we finished moving some things to the sync module. The constant change for consistency

ferdyrod (approved):

andresmr (approved):

PR #5172, opened 2026-09-18
feat: [ANDROAPP-7154] add SyncErrorLogViewModel and its Koin wiring

Balcan:
Adds the ViewModel backing the new sync error log dialog (loading, selection mode, share), the GetSharedData use case it depends on, and the Koin module registration. Not yet reachable from any UI.

## Description


app/src/main/java/org/dhis2/usescases/settings/ui/viewmodels/SyncErrorLogViewModel.kt
```diff
@@ -0,0 +1,95 @@
...
+    private val _shareEvent = MutableSharedFlow<String>()
+    val shareEvent = _shareEvent.asSharedFlow()
+
+    val errorLogUiState =
+        _errorLogUiState
+            .onStart {
+                loadErrorLog()
+            }.stateIn(
+                scope = viewModelScope,
+                started = SharingStarted.WhileSubscribed(5_000),
```
andresmr, line 35: don't we have a shared constant for this hardcoded value?. do you think it would make sense?

app/src/main/java/org/dhis2/usescases/settings/ui/viewmodels/SyncErrorLogViewModel.kt
```diff
@@ -0,0 +1,95 @@
...
+            _errorLogUiState.update {
+                SyncErrorLogUiState.Log(
+                    errorList = result.getOrDefault(emptyList()),
+                )
+            }
+        }
+    }
+
+    fun initSelectionMode() {
+        launchUseCase(dispatchers.io()) {
```
andresmr, line 51: is this just updating the state? does it needs to be moved to an I/O thread? could it be performed in UI?

app/src/main/java/org/dhis2/usescases/settings/ui/viewmodels/SyncErrorLogViewModel.kt
```diff
@@ -0,0 +1,95 @@
...
+            _errorLogUiState.update {
+                SyncErrorLogUiState.Selection(errorList = (it as SyncErrorLogUiState.Log).errorList)
+            }
+        }
+    }
+
+    fun exitSelectionMode() {
+        launchUseCase(dispatchers.io()) {
```
andresmr, line 59: same comment as line 51. Could it be UI thread?

app/src/main/java/org/dhis2/usescases/settings/ui/viewmodels/SyncErrorLogViewModel.kt
```diff
@@ -0,0 +1,95 @@
...
+                        (it as SyncErrorLogUiState.Selection).errorList.map { errorModel ->
+                            errorModel.copy(isSelected = false)
+                        },
+                )
+            }
+        }
+    }
+
+    fun setSelected(selectedIndex: Int) {
+        launchUseCase(dispatchers.io()) {
```
andresmr, line 72: States updates could be done in a UI thread

app/src/main/java/org/dhis2/usescases/settings/ui/viewmodels/SyncErrorLogViewModel.kt
```diff
@@ -0,0 +1,95 @@
...
+                            model
+                        }
+                    }
+                SyncErrorLogUiState.Selection(errorList = updatedList)
+            }
+        }
+    }
+
+    fun onShareLog() {
+        launchUseCase(dispatchers.io()) {
```
andresmr, line 88: same comment ☺️

app/src/main/java/org/dhis2/usescases/settings/ui/viewmodels/SyncErrorLogViewModel.kt
andresmr, line 1: Is the test of this class in another PR?

andresmr (changes requested):

